PR #23, opened 2016-01-22
Fixing the file image provider implementation

david-caro:
Now you if you specify a url and no path in the disk spec, it will
download to the base of the prefix, but if you do, it will move it to
the given path.

Change-Id: I89bc01d322a9ceda8a9de220c957b343e8bebd37
Signed-off-by: Tolik Litovsky <tlitovsk@redhat.com>


david-caro:
Same as #28

PR #40, opened 2016-01-26
Fixes to default domain: allow HAP, move to host-model, disable bootmenu

mykaul:


david-caro:
can you squash the extra merge commit?

Also maybe @tlitovsk want's to take a look as this reverts one of he's changes

mykaul:
Still struggling with GitHub (honestly don't see it as better / simpler
than Gerrit, but perhaps it's just me!)

david-caro:
yep, it's a bit different, I'm also struggling a bit with it. Also a bit disappointing that only the owner of the branch can submit commits to it (unless you are using the project's branch, where only the devels of the project can participate)

PR #74, opened 2016-02-02
Using a more jenkins friendly download progress

### LINKED ISSUE #73: Make download progress jenkins-friendly
Right now it rewrites the whole line on each progress step, that makes jenkins output a whole new line for each refresh (as it does not remove lines), so something similar to repoman would be nicer:

    length 46K ...
    %[-----------------------25------------------------50-----------------------75------------------------]
    %[====================================================================================================]

david-caro:
As it's just updating a progress bar instead of removing the whole line,
in jenkins it's shown as two lines instead of one for each update,
making the logs easier to read.

Fixes #73

Change-Id: I08cc8fb494603772d4cdf99e46d75586a7a481ab
Signed-off-by: David Caro dcaroest@redhat.com

### DIFF
--- a/automation/check-merged.sh
+++ b/automation/check-merged.sh
@@ -2,6 +2,12 @@
 EXPORTED_DIR="$PWD/exported-artifacts"
 DOCS_DIR="$EXPORTED_DIR/docs"
 
+if hash dnf &>/dev/null; then
+    YUM="dnf"
+else
+    YUM="yum"
+fi
+
 
 [[ -d "$EXPORTED_DIR" ]] || mkdir -p "$EXPORTED_DIR"
 
@@ -14,8 +20,7 @@ make docs
 mv docs/_build "$DOCS_DIR"
 
 # Style and unit test
-# required for the formatting
-pip install yapf
+pip install -r tests-requires.txt
 make check-local
 
 echo '~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~'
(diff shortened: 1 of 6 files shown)


automation/check-merged.sh
```diff
@@ -2,6 +2,12 @@
 EXPORTED_DIR="$PWD/exported-artifacts"
 DOCS_DIR="$EXPORTED_DIR/docs"
 
+if hash dnf &>/dev/null; then
```
mykaul: Perhaps 'which dnf' is good too - not sure what the difference is really.

automation/check-patch.sh
```diff
@@ -29,8 +36,7 @@ if ! code_changed; then
 fi
 
 # Style and unit tests
-# required for the formatting
-pip install yapf
+pip install -r tests-requires.txt
```
mykaul: Does it work without root?

automation/check-merged.sh
david-caro: Well, hash is a bash builtin while which is a separated command, hash will ignore aliases, and reads entries from the hash table and allows managing it, while which will not use hash tables but simulate bash behavior, and do the path searching each time.

In this case, you can just swap one with the other.

automation/check-patch.sh
david-caro: depends on your setup, if you are running in a virtualenv yes, it will work without root, but this is meant to be run with mock_runner.sh, and in that case, it's always running as root

nvgoldin:
closing for now, feel free to open if needed.

PR #77, opened 2016-02-03
Ova ovirt support

tlitovsk:
This patch will allow the ovirt contrib plugin to handle the ova disk properly
It will be able to find the correct rootfs in the correct disk order


tlitovsk:
ci please test

david-caro:
... You just merged two dummy commits

PR #78, opened 2016-02-03
Adding use of backing mode for qcow2 files

tlitovsk:
This code will allow 2 things
Use of backing files for qcow2 mode
Allow use of two machines with one original backing file

Change-Id: I173030466ee5b90a1f2613e802086a0d6f1b9307
Signed-off-by: Tolik Litovsky <tlitovsk@redhat.com>

### DIFF
--- a/lib/lago/__init__.py
+++ b/lib/lago/__init__.py
@@ -464,46 +464,66 @@ class Prefix(object):
 
             disk_filename = '%s_%s.%s' % (name, spec['name'], spec['format'])
             disk_path = self.paths.images(disk_filename)
+
             if spec['type'] == 'template':
-                if template_store is None or template_repo is None:
-                    raise RuntimeError('No templates directory provided')
+                template_type = spec.get('template_type', 'lago')
+                if template_type == 'lago':
+                    if template_store is None or template_repo is None:
+                        raise RuntimeError('No templates directory provided')
 
-                template = template_repo.get_by_name(spec['template_name'])
-                template_version = template.get_version(
-                    spec.get('template_version', None)
-                )
-
-                if template_version not in template_store:
-                    LOGGER.info(
-                        log_utils.log_always(
-                            "Template %s not in cache, downloading"
-                        ) % template_version.name,
+                    template = template_repo.get_by_name(spec['template_name'])
+                    template_version = template.get_version(
+                        spec.get('template_version', None)
                     )
-                    template_store.download(template_version)
-                template_store.mark_used(template_version, self.paths.uuid())
 
-                disk_metadata.update(
-                    template_store.get_stored_metadata(
-                        template_version,
-                    ),
-                )
+                    if template_version not in template_store:
+                        LOGGER.info(
+                            log_utils.log_always(
+                                "Template %s not in cache, downloading"
+                            ) % template_version.name,
+                        )
+                        template_store.download(template_version)
+                    template_store.mark_used(
+                        template_version, self.paths.uuid()
+                    )
 
-                base = template_store.get_path(template_version)
-                qemu_img_cmd = [
-                    'qemu-img', 'create', '-f', 'qcow2', '-b', base, disk_path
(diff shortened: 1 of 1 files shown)


tlitovsk:
ci please test

david-caro:
I don't agree with this, it's changing the way it was working before, it should be a different provider instead of file. And I strongly believe that we should formalize the image provider plugin before adding more options

tlitovsk:
Actually it aligns with how the qcow files are used across the system
In template the qcow files used this way.

But weI dont want to make image provider plugin . this is exactly what file provider does.
Lets talk on this on the scrub

tlitovsk:
ci please test

tlitovsk:
moved the option to be part of template

david-caro:
@tlitovsk take a look at http://lago.readthedocs.org/en/latest/CI.html for the ci process, if you are updating the pull request, you don't need to write any comments, only when you want to rerun the tests, or start the merge pipeline

lib/lago/__init__.py
```diff
@@ -464,46 +465,67 @@ def _create_disk(
 
             disk_filename = '%s_%s.%s' % (name, spec['name'], spec['format'])
             disk_path = self.paths.images(disk_filename)
+
             if spec['type'] == 'template':
-                if template_store is None or template_repo is None:
-                    raise RuntimeError('No templates directory provided')
+                template_type = spec.get('template_type', 'repository')
+                if template_type == 'repository':
```
david-caro: better use lago as template type here, repository is too broad (virt-builder uses also a repo, vagrant too, glance is an image repo too)

lib/lago/__init__.py
```diff
@@ -464,46 +465,67 @@ def _create_disk(
...
-                    'qemu-img', 'create', '-f', 'qcow2', '-b', base, disk_path
-                ]
+                    base = template_store.get_path(template_version)
+                    qemu_img_cmd = [
+                        'qemu-img', 'create',
+                        '-f', 'qcow2',
+                        '-b', base, disk_path
+                    ]
 
+                elif template_type == 'file':
```
david-caro: Probably qcow is better here too, or snapshot, file is too broad and can be easily confused with the file image provider

lib/lago/__init__.py
```diff
@@ -620,8 +643,8 @@ def _ova_to_spec(self, filename):
 
         if image_file is not None:
             disk_meta = {"root-partition": "/dev/sda1"}
-
-            disk_spec = [{"type": "file",
+            disk_spec = [{"type": "template",
+                          "template_type": "file",
```
david-caro: this is a backwards incompatible change, you pass from using a  file directly to using a template, I don't think that noone but you is currently using it so if it's ok for you it's ok for me, but keep in mind that it will not be allowed to be that simple to change like that in the future

tlitovsk: this doesn't change anything from the outside world.
neither the user see any change in behaviour.
this file is a temporary file used by lago.
I can even delete it and the user will / should not be aware of it.

david-caro: That's not true, before you had a qcow2 file, that you used directly, and modified it in place, now you create a new one each time with that old one as backing file so you end up with two+ qcows, and not modifying the first one that is extracted from the ova file.
Afaik the file is not removed at the end of the run, and it might be reused (it will only redownload if it's not there), so it's not temporary right?

tlitovsk: This is a temporary file in prefix.
Exactly the same as files inside images directory
its removed together with prefix

david-caro: If you specify it twice for two different hosts (for example as shared disk) you get messed up

david-caro:
It might not be very relevant right now, but it's very important that we maintain a 100% backwards compatibility in the same major version in the future to avoid painful updates

tlitovsk:
ci merge please

PR #84, opened 2016-02-03
Fix changed file detection

david-caro:
Now it compares the current commit with it's first parent, instead of
just looking for changes in the current commit (that worked when not
using merge commits).

Change-Id: I9614bec1e931c35fbf9298db69cc6c58ffa13b9f


david-caro:
ci merge please

PR #95, opened 2016-02-05
Catch (and raise) exception if download of template fails

mykaul:
I'm getting a timeout (of trying to download from the office a master image), but it's not handled at all right now. This code catches it and prints the URL.


mykaul:
#141 takes care of it.

PR #97, opened 2016-02-09
Fix typos, capitalization and add temp dirs to gitignore

mykaul:
- There were several typos all over fixed in this patch
- Renamed some 'Params' to 'Args' for proper docstring
- Add two temp dirs to gitignore


contrib/ovirt/lib/ovirtlago/testlib.py
```diff
@@ -138,7 +138,7 @@ def addError(self, test, err):
     def addFailure(self, test, err):
         self._addFault(test, err)
 
-    def _addFault(self, test, err):
```
david-caro: was this intended?

lib/lago/__init__.py
```diff
@@ -67,16 +67,16 @@ def _ip_in_subnet(subnet, ip):
...
 
     Returns:
-        bool: ``True`` if ip is in subnet, ``False`` otherwise
+        bool: ``True`` if IP is in subnet, ``False`` otherwise
     """
     return (
-        _create_ip(subnet, 1) == _create_ip(ip, 1)
-        or '0.0.0.1' == _create_ip(ip, 1)
+        _create_ip(subnet, 1) == _create_ip(ip, 1) or
+        '0.0.0.1' == _create_ip(ip, 1)
```
david-caro: this might break yapf

contrib/ovirt/lib/ovirtlago/testlib.py
mykaul: Yes, I thought it was a typo, then noticed it was simply not used, so deleted it. I kinda twisted the term 'cleanup' here ;-)

lib/lago/__init__.py
mykaul: Interesting, because pycharms complained before and was happy with this fix.

contrib/ovirt/lib/ovirtlago/testlib.py
david-caro: it's used just in the line before the def, you should change it there too (and any other place it's called if any)

lib/lago/__init__.py
david-caro: We are not using pycharms, there's a pull request on how to set up the dev environment, will try to push it

mykaul:
Hmm, trying to fix the silly yapf issue by following the recommendation to run:
"/usr/bin/python /usr/bin/yapf --style .style.yapf --in-place --recursive ." only makes it go 100% CPU and my CPU to heat up...

david-caro:
Were you able to get it running? If you have a venv or similar it might also recurse it, maybe I should change the docs to use the pre-commit hook instead (will parse only the modified files staged for commit), though that will not run on the files you already commited

contrib/ovirt/lib/ovirtlago/testlib.py
```diff
@@ -132,13 +132,13 @@ def options(self, parser, env=None):
     def configure(self, options, conf):
         super(LogCollectorPlugin, self).configure(options, conf)
 
-    def addError(self, test, err):
-        self._addFault(test, err)
+    def addError(self, test):
```
david-caro: This is part of the nose plugin interface, it might be called by nose with that extra err parameter :/

    http://nose.readthedocs.org/en/latest/plugins/interface.html#nose.plugins.base.IPluginInterface.addError

You might remove it from _addFault though, there's no need for it there

PR #115, opened 2016-02-13
DNS and IPv6 additions to VM networking

mykaul:
Added both DNS and IPv6 to the VM networking
DNS uses lago.example.com as a suffix.
IPv6 uses fd8f:1391:3a82:5e0d:: as subnet (private range)

### DIFF
--- a/lib/lago/net_nat_template.xml
+++ b/lib/lago/net_nat_template.xml
@@ -6,5 +6,8 @@
     </nat>
   </forward>
   <bridge name='@BR_NAME@' stp='on' delay='0' />
+  <dns forwardPlainNames='yes' >
+  </dns>
   <ip address='@GW_ADDR@' netmask='255.255.255.0' />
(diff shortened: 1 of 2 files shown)


mykaul:
Tested by changing the iSCSI server to support IPv6:
...

targetcli iscsi/iqn.2014-07.org.ovirt:storage/tpg1/portals delete 0.0.0.0 3260
targetcli iscsi/iqn.2014-07.org.ovirt:storage/tpg1/portals create ::0
...
iscsiadm -m discovery -t sendtargets -p ::1
...

And it seems to work:
[root@lago_basic_suite_3_6_storage-iscsi ~]# iscsiadm -m session
tcp: [1] [::1]:3260,1 iqn.2014-07.org.ovirt:storage (non-flash)

mykaul:
Forgot the DNS stuff. ARGH.

PR #116, opened 2016-02-13
IPv6 + DNS resolving for VM networking

mykaul:
Added both IPv6 and DNS (.lago.example.com) resolving using libvirt to VMs.
- IPv6 needs small changes to the ifcfg-* of the VMs (or 'dhclient -6') but works otherwise.
[root@lago_basic_suite_3_6_storage-iscsi ~]# hostname -I
192.168.200.4 fd8f:1391:3a82:5e0d::c0a8:c8f0 
[root@lago_basic_suite_3_6_storage-iscsi ~]# ping6 `hostname`
PING lago_basic_suite_3_6_storage-iscsi(lago_basic_suite_3_6_storage-iscsi) 56 data bytes
64 bytes from lago_basic_suite_3_6_storage-iscsi: icmp_seq=1 ttl=64 time=0.008 ms
64 bytes from lago_basic_suite_3_6_storage-iscsi: icmp_seq=2 ttl=64 time=0.021 ms
^C
--- lago_basic_suite_3_6_storage-iscsi ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1000ms
rtt min/avg/max/mdev = 0.008/0.014/0.021/0.007 ms
[root@lago_basic_suite_3_6_storage-iscsi ~]# ping `hostname`
PING lago_basic_suite_3_6_storage-iscsi (192.168.200.4) 56(84) bytes of data.
64 bytes from lago_basic_suite_3_6_storage-iscsi.lago.example.com (192.168.200.4): icmp_seq=1 ttl=64 time=0.028 ms


ifireball:
Id rather have the DNS suffix be configurable for a couple of reasons:
- The default behaviour should stay as is - e.g. short-name resolution without long names.
- There could be multiple networks with managed DNS/DHCP in the same Lago environment, we would like to be able to separate them by DNS suffix.

mykaul:
@david-caro  - any reason this cannot be merged?

david-caro:
There's @ifireball's unanswered comment, I was waiting for a response

mykaul:
Both of them were answered - I hope:
1. There is no reason the short name won't work as it did before - I did not touch that part. In fact, being in the same domain should make it work.
2. The domain name is configurable in the .json.in file. If it does not exist, only then 'lago.example.com' is used.

david-caro:
Sorry, I see no response un the pull request, will take a look

ifireball:
Since you make the default be suffix be '.lago.example.com' instead of nothing as it is now, you are changing current behaviour instead of just adding a feature, for example, right now one can assume for now that backwards-resolving a VM IP in Lago would result in the VM name that can be used, for example, as the 1st parameter to 'lago shell'. When domain suffixes are defined this will no longer be the case.
To maintain compatibility, domain suffixes should be added if and only if requested explicitly in the JSON file.

mykaul:
@ifireball  - OK, latest patch above hopefully is good - if the config isn't there (in the JSON file), DNS isn't configured.
(BTW, to my disappoint, it did not cut a lot from the testing time. I  have hoped it will).

david-caro:
btw. can you please rebase? there are a lot of commits that only pollute the history of the patch

mykaul:
Failing to rebase as I have some unexplained conflicts I fail to resolve. Trying again...

mykaul:
@david-caro  - is it OK now? I've had some unexplained conflicts that I hope I've resolved. Not sure still :(

david-caro:
@mykaul not really, there are still 23 commits, most of them repeated (you can try rebase -i and squash them together) and a couple merge commits (no need for these, they can be squashed into the first commit as if it was rebased instead)

mykaul:
@david-caro  those are my poor attempts to rebase. I'll see what I can do.

david-caro:
:+1:

mykaul:
Sorry, giving up. Enough is enough. Tired of trying to squash this. Will contribute this some other time.

PR #138, opened 2016-02-24
Use a default prefix

david-caro:
This allows not specifying the prefix on init, that will create it at $PWD/.lago, and if that .lago prefix exists, then it will be discovered when running any command on any directory that shares the same parent with it, similar to the git command


david-caro:
ci merge please

ifireball:
I dont see how this can be used in practice given that the optional prefix argument is followed by a non optional config argument. is a single string passed to init now considered to be the config file?

david-caro:
Yep:

```
usage: lago init [-h] [--template-store TEMPLATE_STORE]
                 [--template-repo-name TEMPLATE_REPO_NAME]
                 [--template-repo-path TEMPLATE_REPO_PATH]
                 [PREFIX] VIRT_CONFIG

positional arguments:
  PREFIX                Prefix directory of the deployment, if none passed, it
                        will use $PWD/.lago
  VIRT_CONFIG           Configuration of resources to deploy

  optional arguments:
  -h, --help            show this help message and exit
  --template-store TEMPLATE_STORE
                        Location to store templates at
  --template-repo-name TEMPLATE_REPO_NAME
                        Name of the repo from the template repos dir
  --template-repo-path TEMPLATE_REPO_PATH
                        Repo file describing the templates
```

david-caro:
As easy as:

    lago init path/to/my/env.json
    lago status
    lago start
    ...

PR #141, opened 2016-02-25
Add two RuntimeError exceptions: add the URL of a failed download.

mykaul:
For templates download as well as repo metadata downloads.


lib/lago/templates.py
```diff
@@ -154,13 +154,18 @@ def open_url(self, url, suffix='', dest=None):
             except RuntimeError:
                 pass
         full_url = posixpath.join(self.baseurl, url) + suffix
-        response = urllib.urlopen(full_url)
-        if response.code >= 300:
+        try:
+            response = urllib.urlopen(full_url)
+            if response.code >= 300:
```
david-caro: shouldn't this raise be out of the try? won't it be re-cached?

lib/lago/templates.py
```diff
@@ -154,13 +154,18 @@ def open_url(self, url, suffix='', dest=None):
...
+                raise RuntimeError(
```
david-caro: shouldn't this raise be out of the try? won't it be re-cached?

david-caro:
If you want me to be able to help you with the pull requests, you should create the branch on the lago-project/lago repo instead of your fork, that way I'll be able to push there too :/

contrib/ovirt/lib/ovirtlago/repoverify.py
```diff
@@ -107,7 +107,13 @@ def fetch_xml(url):
     Returns:
         lxml.etree._Element: Root of the xml tree for the retrieved resource
     """
-    content = urllib2.urlopen(url).read()
+    try:
+        content = urllib2.urlopen(url).read()
+    except urllib2.URLError, e:
```
david-caro: you should use the newer recommended form:

    except urllib2.URLError as error

contrib/ovirt/lib/ovirtlago/repoverify.py
```diff
@@ -107,7 +107,13 @@ def fetch_xml(url):
...
+        raise RuntimeError(
+            'Failed to open URL %s\nError code: %s\nError reason: %s' %
+            (url, e.code, e.reason)
```
david-caro: a urllib2.URLError instance has no code member:
  https://docs.python.org/2/library/urllib2.html#urllib2.URLError
But the HTTPError might:
  https://docs.python.org/2/library/urllib2.html#urllib2.HTTPError

lib/lago/templates.py
```diff
@@ -180,7 +185,13 @@ def report(count, block_size, total_size):
 
         if dest:
             response.close()
-            urllib.urlretrieve(full_url, dest, report)
+            try:
+                urllib.urlretrieve(full_url, dest, report)
+            except IOError, e:
```
david-caro: Urlretrieve might also raise a **ContentTooShortError**
https://docs.python.org/2/library/urllib.html#urllib.ContentTooShortError

PR #152, opened 2016-03-01
Added --prefix-path option to the lago command

david-caro:
This allows to explicitly override the automatic discovery of the
prefix path it should be working on

Change-Id: If440b6b5d7d8cf457abe7bf4727fa405fc692c1a


david-caro:
ci :shipit:

PR #166, opened 2016-03-07
Added ignore-warnings option

david-caro:
Also moved to use the python stdlib warnings module, that allows us to
use deprecation warnings and handle them properly.

Change-Id: Ib2891696d6c04b2d9f5567413b769d94c46db7c9


david-caro:
ci :shipit:

PR #169, opened 2016-03-09
Ignore zero sized RPM files when merging repos.

mykaul:
For some reason (unsuccesful downloads?) I got empty RPM files.
They were merged with other (good) files and I had those as RPMs.
Obviously, they failed to install.
The search for the RPMs now ignores such files and thus merges only REAL files.

### DIFF
--- a/ovirtlago/merge_repos.py
+++ b/ovirtlago/merge_repos.py
@@ -67,6 +67,8 @@ def merge(output_dir, input_dirs):
                     input_dir,
                     '-type',
                     'f',
+                    '-size',
+                    '+0',
                     '-name',
                     '*.rpm',
                 ]


david-caro:
Shouldn't it complain instead?

mykaul:
Complaining is a second fix - if reposync did not succeed, we are not very vocal about it.
This one does no harm right now, and does allow me to continue and try and get master installed and running.

david-caro:
I say because you will not be getting all the rpms you expect in the internal_repo (you might be missing some)

david-caro:
I can't do nothing on this pull request, as the branch resides on your fork

mykaul:
I actually will - the problem was that I got the same RPM from two sources
- one was zero sized and the other was the correct one - and the zero sized
remained.
Fixing this allowed me to keep the good one.

The fix does not harm in any case - there is no use for a zero sized RPM.

I agree that in parallel we need to see all downloads succeed and find it
why the download fails.

david-caro:
ci merge please

david-caro:
opened #170

mykaul:
@david-caro any idea what the above failure is all about? Jenkins mis-behaving?

david-caro:
Nop, it actually failed the functional tests:

    not ok 30 ovirt.collect: collect
    # DIFF:Checking if the output differs from the expected
    # CURRENT(<): output | EXPECTED(>): /home/jenkins/workspace/lago_master_check-merged-fc22-x86_64/lago/tests/functional/fixtures/ovirt.collect/output/lago_functional_tests_host/_var_log_vdsm/fancylog2.log
    # 0a1
    # > fancy log number 2

david-caro:
does not seem related at all though

david-caro:
I'm trying to reproduce locally see what's the issue

david-caro:
ci :shipit:

Issue #13, opened 2016-01-21
Move the collect and test verbs to the main cli too


david-caro:
As they would be useful out of ovirt too, was being worked on here: https://gerrit.ovirt.org/#/c/45433/

Issue #12, opened 2016-01-21
collect: allow specifying the artifacts in the vm json


david-caro:
Was being worked here https://gerrit.ovirt.org/#/c/45434/

PR #181, opened 2016-03-15
Collect /var/log/messages from hosts. Useful at times.

mykaul:


mykaul:
Tested with oVirt systems tests against 3.6 - seems to be working (especially as the tests was failing, it was helpful to get the /var/log/messages file!)

david-caro:
Related to #12 and #13

david-caro:
ci merge please

PR #201, opened 2016-03-31
Simplified merge repos

mykaul:
KISS method of merging the repos. Just copying all files per version.
Excluding 686 arch, nothing fancy.

### DIFF
--- a/ovirtlago/merge_repos.py
+++ b/ovirtlago/merge_repos.py
@@ -23,9 +23,6 @@ import os
 import shutil
 from functools import partial
 
-import rpmUtils.arch
-import rpmUtils.miscutils
-
 import utils
 from lago import log_utils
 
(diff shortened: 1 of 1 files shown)


mykaul:
@david-caro  - the failure on EL7 has (hopefully) nothing to do with my patch:
libvirt: QEMU Driver error : monitor socket did not show up: No such file or directory

david-caro:
That's a known issue of libvirt yes

david-caro:
I like the simplification, still have to test it though

david-caro:
Not really, just wanted to give the ovirt system tests a run with it to make sure it does not break anything, but as they are not yet stable I'm focused on fixing them instead. If you verified it, feel free to trigger the merge

david-caro:
heheh :), you should do the merges though the ci process (though it's not working now due to a jenkins bug...) as that also runs the tests as gate (avoiding any errors from getting in) and deploys the newly built version to the stapshot repo.

Now it's ok, but you should really get used to using the ci

PR #205, opened 2016-04-01
Differentiate alive and defined VM

### LINKED ISSUE #203: Speed up lago functional tests by skipping following tests if the start failed
Right now if the vms are started in 'stopped' status by qemu/libvirt, each tests will try to ssh to them and eventually time out, we should make sure in lago that the vm is in 'running' state when trying to ssh to it.

david-caro:
Alive means that it's running, defined that it's defined in libvirt,
but on any other state.

Fixes #203

Change-Id: I1b1d1e06e67eb6ed2b978fc3762c7b4096868a4e

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -325,7 +325,9 @@ def do_shell(prefix, host, args=None, **kwargs):
         raise
 
     if not host.alive():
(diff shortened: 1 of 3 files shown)


david-caro:
ci :shipit:

PR #211, opened 2016-04-10
Few fixes to virt.py

mykaul:
1. Copy artificats from live VMs only via SCP:
- This substantially improve artifacts collection time.
2. Add discard and native IO attributes to the QEMU disk driver
3. Move some SSH initialization code from within loops
4. Switch video to QXL (supposed to be better)
5. Switch cdrom form IDE to SCSI (supposed to be better)
6. Disabled bootmenu

Tested with 3.6 and master.

### DIFF
--- a/lago/virt.py
+++ b/lago/virt.py
@@ -33,7 +33,7 @@ import guestfs
 import libvirt
 import lxml.etree
 import paramiko
-from scp import SCPClient
+from scp import SCPClient, SCPException
 
 import config
 import brctl
@@ -618,14 +618,15 @@ class VM(object):
         return spec
 
     def _open_ssh_client(self):
+        client = paramiko.SSHClient()
+        client.set_missing_host_key_policy(paramiko.AutoAddPolicy(), )
         while self._ssh_client is None:
             try:
-                client = paramiko.SSHClient()
-                client.set_missing_host_key_policy(paramiko.AutoAddPolicy(), )
                 client.connect(
                     self.ip(),
                     username='root',
                     key_filename=self._env.prefix.paths.ssh_id_rsa(),
+                    look_for_keys=False,
                     timeout=1,
                 )
                 return client
@@ -658,15 +659,16 @@ class VM(object):
         ssh_timeout = int(config.get('ssh_timeout'))
         ssh_tries = int(config.get('ssh_tries'))
         start_time = time.time()
+        client = paramiko.SSHClient()
+        client.set_missing_host_key_policy(paramiko.AutoAddPolicy(), )
         while ssh_tries > 0:
             try:
                 LOGGER.debug('still got %d tries' % ssh_tries)
-                client = paramiko.SSHClient()
-                client.set_missing_host_key_policy(paramiko.AutoAddPolicy(), )
                 client.connect(
                     self.ip(),
                     username='root',
                     key_filename=self._env.prefix.paths.ssh_id_rsa(),
+                    look_for_keys=False,
                     timeout=ssh_timeout,
                 )
                 return client
@@ -780,11 +782,14 @@ class VM(object):
 
     def copy_from(self, remote_path, local_path, recursive=True):
         with self._scp() as scp:
-            scp.get(
-                recursive=recursive,
-                remote_path=remote_path,
-                local_path=local_path,
-            )
+            try:
+                scp.get(
+                    recursive=recursive,
+                    remote_path=remote_path,
+                    local_path=local_path,
+                )
+            except SCPException:
+                LOGGER.debug('Remote path %s not found' % remote_path)
 
     @property
     def metadata(self):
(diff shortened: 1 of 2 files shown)


mykaul:
ci

mykaul:
ci :shipit:

mykaul:
ci test please

mykaul:
@david-caro  - can you take a look at CI failures? Not sure I'm at fault here:
ok 37 start.1vm_bridged: start everything at once
ok 38 start.1vm_bridged: start is reentrant
ok 39 start: teardown
ok 40 collect: setup
not ok 41 collect: fails if files to collect don't exist
# (from function `helpers.different' in file tests/functional/helpers.bash, line 85,
#  from function `helpers.run_nook' in file tests/functional/helpers.bash, line 68,
#  in test file tests/functional/collect.bats, line 38)
#   `helpers.run_nook "$LAGOCLI" collect --output "$outdir"' failed
# ~/tests/functional/fixtures/collect ~
# RUNNING:lago collect --output /home/jenkins/workspace/lago_master_check-patch-el7-x86_64/lago/tests/functional/fixtures/collect/output
# --output--
# [33mcurrent session does not belong to lago group.[0m
# [36m@ Collect artifacts: [0m[0m[0m
# [36m  # [Thread-1] vm01: [0m[0m[0m
# [36m  # [Thread-1] vm01: [32mSuccess[0m (in 0:00:09)[0m
# [36m@ Collect artifacts: [32mSuccess[0m (in 0:00:09)[0m
# ---
# "0" != "0"

david-caro:
The issue is that on the collect negative test (to make sure that it fails when it should) it is not failing when the files it tries to collect do not exist

lago/virt.py
```diff
@@ -780,11 +782,14 @@ def copy_to(self, local_path, remote_path):
...
-                local_path=local_path,
-            )
+            try:
+                scp.get(
+                    recursive=recursive,
+                    remote_path=remote_path,
+                    local_path=local_path,
+                )
+            except SCPException:
+                LOGGER.debug('Remote path %s not found' % remote_path)
```
david-caro: Here is where it should rerise the exception to break the execution

lago/virt.py
```diff
@@ -964,6 +967,10 @@ def _create_live_snapshot(self, name):
                     )
                 )
 
+            self.wait_for_ssh()
+            self.guest_agent().start()
+            self.ssh('sync'.split(' '))
+
```
david-caro: What is the reasoning behind moving this?

david-caro:
Btw. small tip on the messages needed to start ci and such, github has now some 'saved replies', I've added there the two ones for the tests and the merge and now I can just hit the 'saved replies' button (on the top right of this same text box, while editing) and select whichever one I want, it's quite useful :smiley:

lago/virt.py
mykaul: It gives few more miliseconds for the logs to be written down, for the SSH to be ready, so 'sync' will do less work here and the SSH is going to wait a bit less. Probably benign.

lago/virt.py
mykaul: I wasn't sure if it should. I'm not sure an empty repo is something we can not continue with - the reason is that those RPMs might be elsewhere - in a different repo - so I opted to continue instead.

david-caro: Then it's a change in the behavior and you should change the tests.
Though it might require some discussion to decide what to do... If I'm collecting the vdsm logs after a successful run, and the logs are not there, I'd want it to fail and not skip them silently.

Probably a better option is to add something like --skip-on-missing or similar and use than in case of failure in the runs to collect the logs.

lago/virt.py
david-caro: okok

mykaul:
@david-caro  - I can't change the SCPException - now after successful ovirt-engine installation, everything blows up as the hosts fail to extract vdsm.log - which is not there yet (it's only test 001 that passed!).

david-caro:
That's why I say that adding a new option to the collect command (--skip-on-missing) might be the best solution, it will be bacjwards compatible to the current code, and allow you to decide if you want to fail on missing logs from the cli.

So when running it just after the engine install, you can jst add the flag (--skip-on-missing) and when collecting after the hosts install, run it without it to make sure to fail if the logs are not there.

mykaul:
@david-caro  in the old implementation, how did it cope with /var/log/vdsm not available after test 001?

except Exception:
                LOGGER.exception(
                    'Failed to copy %s from %s',
                    guest_path,
                    self.name(),


Is that what would have been triggered? Because I don't want to wish anything more to the code - but retrain the same behavior.

david-caro:
Before it failed, not sure how did it skip it on the ovirt-system-tests themselves, but the functional tests that this pr is failing on is actually testing that, the behavior when collect does not find a log.

PR #227, opened 2016-04-27
Fixed wrong error log on reposetup

david-caro:
If the repos dir already existed, the reposetup log would show error
because the exception was not being cleared from the sys.exc_info
stack before starting the log task

Change-Id: Icb599df0d96f1efe330b824a973bc96e2dccf693


david-caro:
ci :shipit:

david-caro:
ci merge please

PR #230, opened 2016-05-03
Add prefix env vars

david-caro:
Now you can use variables on the deploy scripts paths, and some vars are added to allow relocatable prefixes


david-caro:
ci :shipit:

PR #231, opened 2016-05-03
Fix el7 dirlock

david-caro:


david-caro:
ci :shipit:

david-caro:
ci :shipit:

mykaul:
@david-caro  - are you sure we need this lock?

david-caro:
Yep, as we don't just run reposync once, but retry and if failed do a check on the downloaded repos consistence we need to lock other lago processes from doing the same (another question is if we should be doing all that process, that I'm not sure about, but I'd leave that for one of the meetings or ml discussion)

david-caro:
ci :shipit:

PR #251, opened 2016-05-13
Allow any pkg source

david-caro:


david-caro:
This is ready to get merged @eedri

Allows to use repoman sources like this when runnig reposetup:
```
lago ovirt reposetup --reposync-yum-config basic_suite_master/reposync-config.repo --custom-source http://jenkins.ovirt.org/job/ovirt-engine_3.6_build-artifacts-el7-x86_64/132/
```
And that will bring in the ovirt rpms from that job (3.6) and use those instead of the ones from master (defined in the reposync-config.repo) as the custom-sources have priority over the reposync one.

david-caro:
ci :shipit:

PR #262, opened 2016-05-23
Process domains in the same consistent order of domains.

mykaul:
In order to process domains in the order they are in the JSON file,
I've changed all relevant dicts to collections.OrderedDicts as well as
loaded the JSON with the proper settings to keep it ordered.

Testing both oVirt 3.6 and master show that now domains are processed
in a consistent manner: engine, storage, host0, host1

### DIFF
--- a/lago/utils.py
+++ b/lago/utils.py
@@ -17,6 +17,7 @@
 #
 # Refer to the README and COPYING files for full details of the license
 #
+import copy
 import Queue
 import collections
 import datetime
@@ -419,6 +420,10 @@ def deepcopy(original_obj):
     """
     if isinstance(original_obj, list):
         return list(deepcopy(item) for item in original_obj)
+    elif isinstance(original_obj, collections.OrderedDict):
+        return collections.OrderedDict(
+            (key, deepcopy(val)) for key, val in original_obj.items()
+        )
     elif isinstance(original_obj, dict):
         return dict((key, deepcopy(val)) for key, val in original_obj.items())
     else:
@@ -437,12 +442,14 @@ def load_virt_stream(virt_fd):
         dict: Loaded virt config
     """
     try:
-        virt_conf = json.load(virt_fd)
+        virt_conf = json.load(
+            virt_fd, object_pairs_hook=collections.OrderedDict
+        )
     except ValueError:
         virt_fd.seek(0)
         virt_conf = yaml.load(virt_fd)
 
-    return deepcopy(virt_conf)
+    return copy.deepcopy(virt_conf)
 
 
 def in_prefix(prefix_class, workdir_class):
--- a/lago/virt.py
+++ b/lago/virt.py
@@ -17,6 +17,7 @@
 #
 # Refer to the README and COPYING files for full details of the license
 #
+import collections
 import functools
 import hashlib
 import json
@@ -86,11 +87,11 @@ class VirtEnv(object):
         with open(self.prefix.paths.uuid(), 'r') as uuid_fd:
             self.uuid = uuid_fd.read().strip()
 
-        self._nets = {}
+        self._nets = collections.OrderedDict()
         for name, spec in net_specs.items():
             self._nets[name] = self._create_net(spec)
 
-        self._vms = {}
+        self._vms = collections.OrderedDict()
         for name, spec in vm_specs.items():
             self._vms[name] = self._create_vm(spec)
 
@@ -235,17 +236,21 @@ class VirtEnv(object):
         virt_path = functools.partial(prefix.paths.prefixed, 'virt')
 
         with open(virt_path('env'), 'r') as f:
-            env_dom = json.load(f)
+            env_dom = json.load(f, object_pairs_hook=collections.OrderedDict)
 
-        net_specs = {}
+        net_specs = collections.OrderedDict()
         for name in env_dom['nets']:
             with open(virt_path('net-%s' % name), 'r') as f:
-                net_specs[name] = json.load(f)
+                net_specs[name] = json.load(
+                    f, object_pairs_hook=collections.OrderedDict
+                )
 
-        vm_specs = {}
+        vm_specs = collections.OrderedDict()
         for name in env_dom['vms']:
             with open(virt_path('vm-%s' % name), 'r') as f:
-                vm_specs[name] = json.load(f)
+                vm_specs[name] = json.load(
+                    f, object_pairs_hook=collections.OrderedDict
+                )
 
         return cls(prefix, vm_specs, net_specs)
 


lago/prefix.py
```diff
@@ -805,6 +805,7 @@ def virt_conf_from_stream(
         Returns:
             None
         """
+        virt_conf = collections.OrderedDict()
```
david-caro: This is effectively doing nothing, as it's overriden on the next line (with whatever load_virt_stream returns)

lago/utils.py
```diff
@@ -533,12 +534,13 @@ def load_virt_stream(virt_fd):
         dict: Loaded virt config
     """
     try:
-        virt_conf = json.load(virt_fd)
+        virt_conf = collections.OrderedDict()
```
david-caro: this also is overridden in the next line

lago/utils.py
```diff
@@ -533,12 +534,13 @@ def load_virt_stream(virt_fd):
...
+        virt_conf = json.load(virt_fd, object_pairs_hook=collections.OrderedDict)
     except ValueError:
         virt_fd.seek(0)
         virt_conf = yaml.load(virt_fd)
```
david-caro: Is there any way to make sure the order is also preserved when loading yaml files? As well, it's the format we are moving everything to

lago/virt.py
```diff
@@ -247,17 +247,20 @@ def from_prefix(cls, prefix):
...
-                net_specs[name] = json.load(f)
+                net_specs[name] = collections.OrderedDict()
+                net_specs[name] = json.load(f, object_pairs_hook=collections.OrderedDict)
 
-        vm_specs = {}
+        vm_specs = collections.OrderedDict()
         for name in env_dom['vms']:
             with open(virt_path('vm-%s' % name), 'r') as f:
-                vm_specs[name] = json.load(f)
+                vm_specs[name] = collections.OrderedDict()
```
david-caro: same as before, overriden

lago/virt.py
```diff
@@ -247,17 +247,20 @@ def from_prefix(cls, prefix):
...
-            env_dom = json.load(f)
+            env_dom = collections.OrderedDict()
+            env_dom = json.load(f, object_pairs_hook=collections.OrderedDict)
 
-        net_specs = {}
+        net_specs = collections.OrderedDict()
         for name in env_dom['nets']:
             with open(virt_path('net-%s' % name), 'r') as f:
-                net_specs[name] = json.load(f)
+                net_specs[name] = collections.OrderedDict()
```
david-caro: same as before, overriden

lago/virt.py
david-caro: This has no effect, is overriden on the next line

lago/virt.py
```diff
@@ -247,17 +247,20 @@ def from_prefix(cls, prefix):
...
+                net_specs[name] = json.load(f, object_pairs_hook=collections.OrderedDict)
```
david-caro: This has no effect, is overriden on the next line

lago/utils.py
david-caro: This has no effect, is overriden on the next line

lago/prefix.py
david-caro: This has no effect, is overriden on the next line

david-caro: This has no effect, is overriden on the next line

lago/utils.py
david-caro: This has no effect, is overriden on the next line

lago/virt.py
```diff
@@ -247,17 +247,20 @@ def from_prefix(cls, prefix):
         virt_path = functools.partial(prefix.paths.prefixed, 'virt')
 
         with open(virt_path('env'), 'r') as f:
-            env_dom = json.load(f)
+            env_dom = collections.OrderedDict()
```
david-caro: This has no effect, is overriden on the next line

lago/virt.py
david-caro: This has no effect, is overriden on the next line

lago/virt.py
david-caro: This has no effect, is overriden on the next line

lago/utils.py
david-caro: Can we make sure that the order is also the same when loaded from yaml? We are using it in more and more places (is nicer to work with manually than json)

david-caro: I seems it's not so trivial though:

http://stackoverflow.com/questions/5121931/in-python-how-can-you-load-yaml-mappings-as-ordereddicts#5121963

mykaul:
ci

lago/utils.py
```diff
@@ -533,12 +534,12 @@ def load_virt_stream(virt_fd):
...
     """
     try:
-        virt_conf = json.load(virt_fd)
+        virt_conf = json.load(virt_fd, object_pairs_hook=collections.OrderedDict)
     except ValueError:
         virt_fd.seek(0)
         virt_conf = yaml.load(virt_fd)
 
-    return deepcopy(virt_conf)
+    return copy.deepcopy(virt_conf)
```
david-caro: this does not replace the old deepcopy function, probably  chose a bad naming back then.

mykaul: Indeed, but the original (copy.deepcopy() function) seems to do the work here.

david-caro: On 05/24 04:34, Yaniv Kaul wrote:
> >      except ValueError:
> >          virt_fd.seek(0)
> >          virt_conf = yaml.load(virt_fd)
> >  
> > -    return deepcopy(virt_conf)
> > +    return copy.deepcopy(virt_conf)
> 
> Indeed, but the original (copy.deepcopy() function) seems to do the work here.

That's because we are not using the yaml anchors[1] anywhere, if you did, it
would break (as the shared dicts and lists are shared as links to the same,
and not independent copies).
That was the only reason to run deepcopy there, as I said, probably the naming
should have been better, something like 'remove_internal_cross_objects' or
something more verbose would have been better.


[1] https://en.wikipedia.org/wiki/YAML#Repeated_nodes
> 
> ---
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/lago-project/lago/pull/262/files/9e9c7b1a9a17511353aeca189f7ebe75ee61ed4a#r64374333

-- 
David Caro

Red Hat S.L.
Continuous Integration Engineer - EMEA ENG Virtualization R&D

Tel.: +420 532 294 605
Email: dcaro@redhat.com
IRC: dcaro|dcaroest@{freenode|oftc|redhat}
Web: www.redhat.com
RHT Global #: 82-62605

mykaul: I added to the deepcopy the ability to copy OrderedDict - but it still did not work for :(
Only when I moved to this implementation, it succeeded.
Any idea why?

david-caro: How did you do that? can you update this pull with that change? (or sent another one, or a pastebin)

mykaul: if isinstance(original_obj, list):
        return list(deepcopy(item) for item in original_obj)
    elif isinstance(original_obj, dict):
        return dict((key, deepcopy(val)) for key, val in
original_obj.items())
    elif isinstance(original_obj, collections.OrderedDict):
        return collections.OrderedDict((key, deepcopy(val)) for key, val in
original_obj.items())
    else:
        return original_obj

david-caro: looking :+1:

david-caro: In that if/elif, you should put the OrderedDict elif before the dict one, as an ordereddict is an instance of dict:
```
In [3]: import collections

In [4]: a = col
%colors      collections  

In [4]: a = collections.OrderedDict()

In [5]: isinstance(a, dict)
Out[5]: True
```

mykaul:
ci

mykaul:
ci

mykaul:
@david-caro  - do you think the above failures are due to my change?
"current session does not belong to lago group." sounds like a suspect?

david-caro:
@mykaul - Yep, I think they are related, not sure exactly how though. The warning about the lago group can be ignored as the tests run as root, has been there since we added the warning.

I can try debugging it tomorrow if you don't have time.

mykaul:
ci

david-caro:
ci test please

It needs the full sentence :)

PR #272, opened 2016-06-01
Added a callback function to handle import exception

gbenhaim:
ci test please

### DIFF
--- a/lago/plugins/__init__.py
+++ b/lago/plugins/__init__.py
@@ -21,6 +21,9 @@
 """
 """
 from stevedore import ExtensionManager
+import logging
+
+LOGGER = logging.getLogger(__name__)
 
 #: Map of plugin type string -> setuptools entry point
 PLUGIN_ENTRY_POINTS = {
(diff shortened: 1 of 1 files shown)


lago/plugins/__init__.py
```diff
@@ -58,7 +61,12 @@ def load_plugins(namespace, instantiate=True):
     Returns:
         dict of str, object: Returns the list of loaded plugins
     """
-    mgr = ExtensionManager(namespace=namespace, )
+    mgr = ExtensionManager(namespace=namespace,
+                           on_load_failure_callback=(
+                               lambda manager, ep, err: LOGGER.debug(
```
david-caro: Usually if you are not going to use one of the params, you set it to '_', like:

`lambda _, ep, err:...`

also, try to be consistent with the indenting (if you are breaking the line after the '(' then do so always).

Though as long as it passes the yapf check I'll merge ;)

david-caro:
you can squash the last two commits in one (git rebase -i origin/master, and change the last from 'pick' to 'f', that will squash it into the previous one)

gbenhaim:
ci :shipit:

david-caro:
ci :shipit:

david-caro:
ci :shipit:

Issue #149, opened 2016-02-29
Add support for chroots and/or containers


david-caro:
Aside from the vms, that will require major refactoring of the virt module to do it right

ifireball:
Can this be a plugin? "lago mock ...", and "lago docker ..." ? I guess we will need ways to create 'lago status' plugins and JSON language plugins.

david-caro:
that's a bit more of an issue, as that is not consistent with the current architecture, as there's no 'lago vm' command (or similar) to support that way of working, we should first support the lago vm one, and that's a full invertion of the current way of working (instead of action+object, you start running object+action, start vm_name vs vm start)

ifireball:
well, I think package repos used to work that way (I don't have the ovirt plugin installed atm so can't check), it will be somewhat more consistent with how git/puppet do things. (git remote, git config, etc.).
OTOH I've no strong opinion here but we may have a name spacing issue (how do you tell it that you meant the 'engine' container and not the 'engine' vm, or maybe names should never collide like this).
I would like it to be possible to add "things" with plugins.

david-caro:
My idea was that just change the backend that provides the 'vm', where vm can be an actual vm, a container, a chroot or whatever, so in that case, the name collision issue is the same as having two vms with the same name, up to you.

The plugins here would provide new backends, not every plugin has to be a command plugin

david-caro:
btw. the 'just' there is an euphemism... :/

eedri:
After talking with @fabiand, maybe we should consider using Kubenetes to actually support containers inside Lago.
So instead of managing containers ourselves on the bare metal, we'll have a VM running Kubenetes and it will manage other nested VMs as PODs which will support running containers.

@fabiand maybe you can elaborate on what we talked about?

ifireball:
This is not a replacement for this issue. We want container support in Lago to allow doing things more efficiently when a full VM is not needed, for example, for letting lago provide services for tests like storage or DNS, etc.

eedri:
But the question is how costly it would be to manage those containers alone and if we have a valid use case to support it, rather than using Kubenetes to do it instead.

ifireball:
Considering nested Kubernetes for this is like considering nested Katello to replace "lago ovirt reposync". It does not make sense at all. We want containers to conserve resources. Kubernetes, for all its benefits, will not do that on the small scale that Lago operates in.

[...]

PR #276, opened 2016-06-06
[Feedback request] Lago docker container support

iankko:
<br/>
DO NOT MERGE YET, THIS IS WORK IN PROGRESS (feedback appreciated)
<br/>

Issue https://github.com/lago-project/lago/issues/149 speaks about RFE to add containers support to Lago. After further inspection, looks we have the following possibilities how this could be implemented (assuming Lago container support assumes the corresponding yaml / json file should be something in the format of existing [docker-compose](https://docs.docker.com/compose/compose-file/) one. IOW the implementation not to allow just one simple Dockerfile, but rather multiple container services like used e.g. in ```docker-compose``` configuration. Thus assuming something like the following format should be accepted by Lago once [Docker] container support is implemented:
  * https://docs.docker.com/compose/django/
  * https://docs.docker.com/compose/rails/
  * https://docs.docker.com/compose/wordpress/

To get to the possibilities as I see them:
* Approach #1 - allow init json / yaml file in the form of ```docker-compose``` file. Use ```docker-compose``` tool internally to handle tasks like ```lago init, lago deploy, lago start, lago stop, etc.```
(IOW lago commands would be just wrapper around internal use of ```docker-compose``` tool, which would be used for parsing the yaml / json file),
  Advantages of this approach:
  * No need to write own ```docker-compose``` yaml/json file parser,
  Disadvantage:
  * Dependency on ```docker-compose``` RPM, which AFAIK  in Fedora (based on v1.2.0) doesn't support ```version 2``` of ```docker-compose.yml``` yet (hopefully we could help with co-maintaining that package and get it updated sooner),

* Approach #2 - allow lago container yaml/json file to have the format of ```docker-compose``` file. But use own parser to extract the information, and perform lago commands (```start, stop, etc```).
  Advantages of this approach:
  * No lago dependency on external (possibly outdated) ```docker-compose``` RPM package,
  Disadvantages of this approach:
  * Need to write own ```docker-compose``` yaml/json file parser. From what I have tested, [python-dockerfile-parse](https://admin.fedoraproject.org/pkgdb/package/rpms/python-dockerfile-parse/) doesn't support parsing of ```docker-compose``` yaml/json files (yet).
  * Following this approach we would basically duplicate the work, ```docker-compose``` is doing.

* Approach #3 - don't allow lago container yaml/json file to have format of ```docker-compose``` file. But define own one, easily to parse.

  Advantages of this approach:
  * We can define the init file format in the way so it will be easy to write the parser for that (we can use [python-docker-py](https://admin.fedoraproject.org/pkgdb/package/rpms/python-docker-py/) functionality to implement containers management for lago (once init file got parsed).

  Disadvantages of this approach:
  * We would diverge from commonly used [docker-compose](https://github.com/search?q=in%3Apath+docker-compose.yml+extension%3Ayml&type=Code) init file. Meaning users would need to get accustomed to a new init file format (different UX than with ```docker-compose``` files).

Feedback / comments on the aforementioned approaches how to get container support for Lago is appreciated.

Thank you, Jan.


david-caro:
I'd start with the vm-provider first, to allow using single docker containers the same way we use vms right now so you can have a mixed env.

Adding the docker compose integration can be done later. I would not couple too tightly with it (that is, I don't think we should use the same format), but I think we should rather allow some kind of 'include' as a part of the environment (running docker-compose as a step to generate the env).

Implementing both as external plugins (as python-ovirtlago), so they are not required for the minimal lago installation

david-caro:
Let's discuss this better on the list :)

iankko:
> Let's discuss this better on the list :)

Did my email actually reach the list? (doesn't seem to have been approved so far)

david-caro:
Yep, it did reach it, and just replied :), any new email will be automatically accepted

iankko:
> I'd start with the vm-provider first, to allow using single docker containers the same way we use vms right now so you can have a mixed env.

Ok, so the request is not to have ```LagoInitFile``` to contain just container definitions, but combined environment ```domains``` + ```containers```. AFAICT this shouldn't require custom Lago prefix provider, just custom vm-provider / container-provider should be sufficient.

> Adding the docker compose integration can be done later. I would not couple too tightly with it (that is, I don't think we should use the same format), but I think we should rather allow some kind of 'include' as a part of the environment (running docker-compose as a step to generate the env).

Let's try to identify the new / enhanced expected format of ```LagoInitFile``` in more detailed way. Would the new container support mean ```LagoInitFile``` in the form e.g. like below [*]:

```
{
  "domains" {
    "server" {... }
    "client_1" {..}
    "client_2" { }
    ...
    "nets" {}
  }
  "containers" {
    "cont_1" {
      "provider" : "docker",
      "format": "dockefile",
      "definition" : "/../path/to/specific/dockerfile"
    }
    "cont_2" {
      "provider" : "rkt",
      "format" : "rktfile",
     "definition" : "/.../path/to/specific/rktfile"
    }
   "cont_set" {
    "provider" : "docker-compose",
    "format" : "docker-compose-ymlfile",
    "definition" : "/path/../to/specific/docker-compose-yml/file"
   }
   "cont_3" {
     "provider" : "vagrant",
     "format" : "vagrantfile",
     "definition" : "/../path/to/specific/vagrantfile"
   }
  }
```

Of course for the beginning (this to be a bit simpler), we would start just with adding support for docker containers and simple Dockerfiles (one Dockerfile per container definition). Just trying to design the format that would be possibly easily expandable in the future with new container types (if / when desired).

Looking at the above proposal, maybe "format" field wouldn't be necessary, if / once "provider" would be required field to supply.

> Implementing both as external plugins (as python-ovirtlago), so they are not required for the minimal lago installation

Right, this is reasonable (so minimal lago install wouldn't require too much RPM dependencies to be able to install). So will count with python-dockerlago and python-dockercomposelago for now.

--

[*} It wouldn't be new format in the sense the old one would stop working (being supported). But new one in the sense of enhanced one.

Thanks, Jan

david-caro:
On 06/08 03:53, Ján Lieskovský wrote:
> > I'd start with the vm-provider first, to allow using single docker containers the same way we use vms right now so you can have a mixed env.
> 
> Ok, so the request is not to have ```LagoInitFile``` to contain just container definitions, but combined environment ```domains``` + ```containers```. AFAICT this shouldn't require custom Lago prefix provider, just custom vm-provider / container-provider should be sufficient.

> > Adding the docker compose integration can be done later. I would not couple too tightly with it (that is, I don't think we should use the same format), but I think we should rather allow some kind of 'include' as a part of the environment (running docker-compose as a step to generate the env).
> 
> Let's try to identify the new / enhanced expected format of ```LagoInitFile``` in more detailed way. Would the new container support mean ```LagoInitFile``` in the form e.g. like below [*]:
> 
> ```
> {
>   "domains" {
>     "server" {... }
>     "client_1" {..}
>     "client_2" { }
>     ...
>     "nets" {}
>   }
>   "containers" {
>     "cont_1" {
>       "provider" : "docker",
>       "format": "dockefile",
>       "definition" : "/../path/to/specific/dockerfile"
>     }
>     "cont_2" {
>       "provider" : "rkt",
>       "format" : "rktfile",
>      "definition" : "/.../path/to/specific/rktfile"
>     }
>    "cont_set" {
>     "provider" : "docker-compose",
>     "format" : "docker-compose-ymlfile",
>     "definition" : "/path/../to/specific/docker-compose-yml/file"
>    }
>    "cont_3" {
>      "provider" : "vagrant",
>      "format" : "vagrantfile",
>      "definition" : "/../path/to/specific/vagrantfile"
>    }
>   }
> ```
> 
> Of course for the beginning (this to be a bit simpler), we would start just with adding support for docker containers and simple Dockerfiles (one Dockerfile per container definition). Just trying to design the format that would be possibly easily expandable in the future with new container types (if / when desired).

I would not distinguish at all between containers and non-containers, just use
the provider for that:

domains:
    server1:
        vm-provider: docker-container
        ...
    server1:
        vm-provider: local-libvirt
    ...

And then for 'includes' something like:

include:
    path/to/docker-compose/file:
        format: docker-compose
    path/to/vargant/file:
        format: vagrant

or if sticking with providers

include:
    path/to/docker-compose/file:
        env-provider: docker-compose
    path/to/vargant/file:
        env-provider: vagrant


Keep in mind that the virtenv is not yet moved to the plugin structure, so in
order to have virtenv plugins changes must be made to lago core code (same for
network providers)

> 
> Looking at the above proposal, maybe "format" field wouldn't be necessary, if / once "provider" would be required field to supply.
> 
> > Implementing both as external plugins (as python-ovirtlago), so they are not required for the minimal lago installation
> 
> Right, this is reasonable (so minimal lago install wouldn't require too much RPM dependencies to be able to install). So will count with python-dockerlago and python-dockercomposelago for now.
> 
> --
> 
> [*} It wouldn't be new format in the sense the old one would stop working (being supported). But new one in the sense of enhanced one.
> 
> Thanks, Jan
> 
> ---
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/lago-project/lago/pull/276#issuecomment-224555393

-- 
David Caro

Red Hat S.L.
Continuous Integration Engineer - EMEA ENG Virtualization R&D

Tel.: +420 532 294 605
Email: dcaro@redhat.com
IRC: dcaro|dcaroest@{freenode|oftc|redhat}
Web: www.redhat.com
RHT Global #: 82-62605

eedri:
ci test please

gbenhaim:
@iankko is there any progress with the patch ?

iankko:
@gbenhaim 
> @iankko is there any progress with the patch ? 

Got stuck due to summer holidays. Gimme couple of days yet.

Thanks, Jan

nvgoldin:
Hey @iankko, I'm closing this for now, if you want to start working on it again feel free to re-open.

PR #282, opened 2016-06-20
Add support for dynamic internal repo path

irosenzw:
Adding support for creating an 'internal repo' folder not in the default location: "deployment-.../default/internal_repo/"

### DIFF
--- a/lago/prefix.py
+++ b/lago/prefix.py
@@ -104,6 +104,14 @@ class Prefix(object):
         self.paths = paths.Paths(self._prefix)
         self._virt_env = None
         self._metadata = None
+        self._internal_repo = None
+
+    def _get_internal_repo (self):
+        return self._internal_repo
+
+    def _set_internal_repo (self, path):
+        self._internal_repo = path
+
 
     def _get_metadata(self):
         """
--- a/lago/ssh.py
+++ b/lago/ssh.py
@@ -354,7 +354,7 @@ def get_ssh_client(
                 )
 
             ssh_tries -= 1
-            time.sleep(1)
+            time.sleep(5)
         else:
             end_time = time.time()
             raise RuntimeError(
--- a/ovirtlago/__init__.py
+++ b/ovirtlago/__init__.py
@@ -210,7 +210,7 @@ class OvirtPrefix(Prefix):
         repo_names,
         custom_sources=None,
         projects_list=None,
-        dist_out=None,
+        internal_repo_path=None,
     ):
 
         if not projects_list:
@@ -220,10 +220,6 @@ class OvirtPrefix(Prefix):
         custom_sources = custom_sources or []
 
         rpm_dirs = []
-
-        dist_output = self.paths.internal_repo()
-        if (dist_out):
-            dist_output = dist_out
 
         for dist in dists:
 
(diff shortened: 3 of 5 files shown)


eedri:
ci test please

lago/prefix.py
```diff
@@ -104,6 +104,14 @@ def __init__(self, prefix):
         self.paths = paths.Paths(self._prefix)
         self._virt_env = None
         self._metadata = None
+        self._internal_repo = None
+
+    def _get_internal_repo (self):
+        return self._internal_repo
+
+    def _set_internal_repo (self, path):
+        self._internal_repo = path
```
david-caro: this is not needed, there's no extra access logic or anything. This is not java ;)

lago/ssh.py
```diff
@@ -354,7 +354,7 @@ def get_ssh_client(
                 )
 
             ssh_tries -= 1
-            time.sleep(1)
+            time.sleep(5)
```
david-caro: seems unrelated?

ovirtlago/__init__.py
```diff
@@ -210,15 +210,24 @@ def _create_rpm_repository(
         repo_names,
         custom_sources=None,
         projects_list=None,
+        dist_out=None,
```
david-caro: this name is misleading, the param that's called 'dists' is for 'distributions', as in fc23/el7. While here dist_out dose seem to be an output path?

ovirtlago/__init__.py
```diff
@@ -210,15 +210,24 @@ def _create_rpm_repository(
...
             projects_list = PROJECTS_LIST
 
+
         custom_sources = custom_sources or []
 
         rpm_dirs = []
+
+        dist_output = self.paths.internal_repo()
+        if (dist_out):
+            dist_output = dist_out
```
david-caro: it's nicer to do both in the same line, probably without creating an extra var, a 'dummy' example:

`dist_out = dist_out or self.paths.internal_repo()`

The above is readable, but does match empty stings and the like (anything python considers false), if you want to allow them you should use:

`dist_out = dist_out if dist_out is not None else self.paths.internal_repo()`

david-caro: Also, I'd put the assignment at the start of the function (or closer to where it's used)

ovirtlago/__init__.py
```diff
@@ -249,8 +259,10 @@ def prepare_repo(
         rpm_repo=None,
         reposync_yum_config=None,
         skip_sync=False,
-        custom_sources=None
+        custom_sources=None,
+        dist_out=None,
     ):
+        _dist_out = dist_out
```
david-caro: this extra assignment seems to do nothing?

ovirtlago/cmd.py
```diff
@@ -114,6 +114,11 @@ def do_ovirt_runtest(prefix, test_file, **kwargs):
     type=os.path.abspath,
 )
 @cli_plugin_add_argument(
+    '--repo-path',
```
david-caro: I'd be a little bit more explicit, something like '--internal-repo-path' or '--prefix-repo-path'. There's another option that has already a bad naming just before '--rpm-repo' that makes this one confusing :confused:

ovirtlago/cmd.py
```diff
@@ -157,6 +162,7 @@ def do_ovirt_reposetup(
     if kwargs['ioprocess_dir']:
         warnings.warn('Deprecated option --ioprocess-dir ignored')
 
+    _repo_path = kwargs['repo_path']
```
david-caro: as this is a non-deprecated option, is better if you declare it as a parameter of the function (just as prefix or rpm_repo), that way you don't need to access the kwargs yourself

david-caro:
I'm curious :smile:, what's your use case?

irosenzw:
Open new PR: https://github.com/lago-project/lago/pull/287

PR #291, opened 2016-07-19
support for custom 'internal_repo' path

irosenzw:
Adding support for creating an 'internal_repo' folder that isn't in the default location: "deployment-.../default/internal_repo/"

### DIFF
--- a/lago/prefix.py
+++ b/lago/prefix.py
@@ -104,6 +104,7 @@ class Prefix(object):
         self.paths = paths.Paths(self._prefix)
         self._virt_env = None
         self._metadata = None
+        self.internal_repo = None
 
     def _get_metadata(self):
         """
--- a/ovirtlago/__init__.py
+++ b/ovirtlago/__init__.py
@@ -210,6 +210,7 @@ class OvirtPrefix(Prefix):
         repo_names,
         custom_sources=None,
         projects_list=None,
+        internal_repo_path=None,
     ):
 
         if not projects_list:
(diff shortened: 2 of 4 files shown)


eedri:
ci test please

lago/prefix.py
```diff
@@ -362,16 +369,14 @@ def _register_preallocated_ips(self, conf):
...
                     raise RuntimeError(
-                        'IP %s was to several domains: %s %s' % (
-                            nic['ip'],
-                            dom_name,
-                            ' '.join(conflict_list),
-                        ),
+                        'IP %s was to several domains: %s %s' %
+                        (nic['ip'],
+                         dom_name,
+                         ' '.join(conflict_list), ),
```
david-caro: This changes are not related to the commit, try not to mix indent/style changes with the commit changes please.

lago/ssh.py
```diff
@@ -354,7 +359,7 @@ def get_ssh_client(
                 )
 
             ssh_tries -= 1
-            time.sleep(1)
+            time.sleep(5)
```
david-caro: This change is totally alien to the pr too, pleas split into another pr (or explain the relation to the custom internal repo)

lago/prefix.py
```diff
@@ -104,6 +104,13 @@ def __init__(self, prefix):
         self.paths = paths.Paths(self._prefix)
         self._virt_env = None
         self._metadata = None
+        self._internal_repo = None
+
+    def _get_internal_repo(self):
+        return self._internal_repo
+
+    def _set_internal_repo(self, path):
```
david-caro: There's no need for getters and setters, this is python, use prefix.internal_repo instead (just make it public by convention)

david-caro:
The style changes you introduced are the ones that are breaking the build, make sure to run yapf on the code you change to reformat to the same style that it's used in the rest of the code:

http://lago.readthedocs.io/en/stable/Env_Setup.html#style-formatting

btw. we are right now pinned on yapf==0.7.1, you can see all the test deps in the test-requires.txt.

mykaul:
@irosenzw  - can you reply to the comments, rebase and repost?

PR #293, opened 2016-08-10
Changing the ssh connection waiting method

irosenzw:
The the current 'wait_for_ssh' method is based on TCP timeout (the default ssh timeout in lago is 10 sec)
when the host is booting, while it doesn't get an IP the timeout is really 10 sec,
but between the time it gets it's IP to the time it opens port 22 the timeout is 0.

So, if one of my vms need more time to open port 22 and start sshd, it might fail after 50 retries.

This fix focus on the maximal 'wait-time' Lago will wait until the host can be reached,
instead of the number of retries, that may vary with different OS.

### DIFF
--- a/lago/plugins/vm.py
+++ b/lago/plugins/vm.py
@@ -360,7 +360,7 @@ class VMPlugin(plugins.Plugin):
         return ssh.wait_for_ssh(
             ip_addr=self.ip(),
             host_name=self.name(),
-            connect_retries=self._spec.get('boot_time_sec', 50),
+            connect_timeout=self._spec.get('boot_time_sec', 600),
             ssh_key=self.virt_env.prefix.paths.ssh_id_rsa(),
             username=self._spec.get('ssh-user'),
             password=self._spec.get('ssh-password'),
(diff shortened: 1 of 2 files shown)


gbenhaim:
@david-caro 
Why do we use retries and timeout for each retry, instead of a global timeout ?
The timeout for each retry is only effective when the host is unreachable and not when sshd isn't yet been started, correct ?

david-caro:
@gbenhaim I don't really remember, I think it's just two different approaches for the same thing. One is by number of tries and the other by timeout. it might be that the first try takes more than the timeout (hopefully not) so that even if you would be able to connect you finally are not, or it might be that 10 retries take way too long and have you waiting for one hour to finally tell you that it failed to connect... right now I think that the former might be a better option, though I have no strong opinion on it.

david-caro:
I guess that the best way to decide is to try both, if this works better, let's go for it.

@irosenzw do you mind rebasing again? (I really hope that github somehow improves that workflow)

gbenhaim:
ci test please

gbenhaim:
@irosenzw can you please fix the code styling ?

gbenhaim:
ci test please

mykaul:
@gbenhaim - any reason not to rebase, test and push?

gbenhaim:
ci test please

gbenhaim:
ci merge please

gbenhaim:
ci merge please

PR #296, opened 2016-08-24
Search for kvm executable too

pilou-:
Not sure if this [fix](https://github.com/lago-project/lago/blob/master/automation/common.sh#L116) is still required.

### DIFF
--- a/lago/vm.py
+++ b/lago/vm.py
@@ -226,13 +226,22 @@ class LocalLibvirtVMProvider(vm.VMProviderPlugin):
         with open(_path_to_xml('dom_template.xml')) as xml_fd:
             dom_raw_xml = xml_fd.read()
 
-        qemu_kvm_path = [
-            path
-            for path in [
-                '/usr/libexec/qemu-kvm',
-                '/usr/bin/qemu-kvm',
-            ] if os.path.exists(path)
-        ].pop()
+        capabilities_raw_xml = self.libvirt_con.getCapabilities()
+        capabilities_xml = lxml.etree.fromstring(capabilities_raw_xml)
+        qemu_kvm_path = capabilities_xml.findtext(
+            "guest[os_type='hvm']/arch[@name='x86_64']/domain[@type='kvm']"
+            "/emulator"
+        )
+
+        if not qemu_kvm_path:
+            LOGGER.warning("hardware acceleration not available")
+            qemu_kvm_path = capabilities_xml.findtext(
+                "guest[os_type='hvm']/arch[@name='x86_64']"
+                "/domain[@type='qemu']/../emulator"
+            )
+
+        if not qemu_kvm_path:
+            raise Exception('kvm executable not found')
 
         replacements = {
             '@NAME@': self._libvirt_name(),
--- a/automation/common.sh
+++ b/automation/common.sh
@@ -113,10 +113,6 @@ run_installation_tests() {
 
 run_basic_functional_tests() {
     local res
-    # Ugly fix to be able to run el* on fc*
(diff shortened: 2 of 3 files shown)


david-caro:
ci test please

david-caro:
I think that the fix would still be required, as the issue happens because the libvirt daemon that we use is running out of the chroot where we run lago for the testing, and the binary inside the chroot is not in the same path as outside.

david-caro:
this does not seem to work well on el7:
```
16:15:47 # libvirt: QEMU Driver error : Cannot check QEMU binary /usr/libexec/qemu-kvm: No such file or directory
```

pilou-:
ci test please

mykaul:
@pilou- can you rebase and we'll retest?

pilou-:
@mykaul done :)

david-caro:
ci test please

pilou-:
Instead of hardcoding path of kvm executable, ask libvirt using [capabilities](https://libvirt.org/formatcaps.html).

I am not sure if the case "Nested virtualization not available" should be handled or not: some functional tests could be run without nested virtualization (but not the whole test suite).

Successfully tested on RHEL 7.2 using `patch-only` test suite.

pilou-:
Could you rerun the tests ?

eedri:
ci test please

lago/vm.py
```diff
@@ -226,13 +226,20 @@ def _libvirt_xml(self):
...
-            path
-            for path in [
-                '/usr/libexec/qemu-kvm',
-                '/usr/bin/qemu-kvm',
-            ] if os.path.exists(path)
-        ].pop()
+        capabilities_raw_xml = self.libvirt_con.getCapabilities()
+        capabilities_xml = lxml.etree.fromstring(capabilities_raw_xml)
+        qemu_kvm_path = capabilities_xml.findtext(
+            "guest[os_type='hvm']/arch/domain[@type='kvm']/emulator"
```
mykaul: Shouldn't the arch be /arch[name='x86_64'] ?

pilou-: Indeed, `x86_64` is hardcoded in `lago/dom_template.xml`: PR updated using `/arch[@name='x86_64']`.

lago/vm.py
```diff
@@ -226,13 +226,22 @@ def _libvirt_xml(self):
...
-        ].pop()
+        capabilities_raw_xml = self.libvirt_con.getCapabilities()
+        capabilities_xml = lxml.etree.fromstring(capabilities_raw_xml)
+        qemu_kvm_path = capabilities_xml.findtext(
+            "guest[os_type='hvm']/arch[@name='x86_64']/domain[@type='kvm']"
+            "/emulator"
+        )
+
+        if not qemu_kvm_path:
+            # Nested virtualization not available
```
gbenhaim: This query can tell you if hardware acceleration is enabled, 
I don't see how this is relates to nested virtualization.
Am I missing something ?

pilou-: If lago is executed inside a VM and if nested virtualization is disabled on the host then this query will return nothing, hence the comment.

But if hardware acceleration is disabled on the host, this query will return nothing too.

I propose to update the comment with: `hardware acceleration (or nested virtualization) not available`, what do you think ?

gbenhaim: I think that we need to keep only ```hardware acceleration not available```, and also print it to the log.

pilou-: @gbenhaim ok, PR updated

gbenhaim:
ci test please

lago/vm.py
gbenhaim: +1

gbenhaim:
ci merge please

PR #302, opened 2016-08-30
Add a restart command to lago/lagocli and implement it for libvirt

karmab:
Add a restart command to lago/lagocli and implement it for libvirt

### DIFF
--- a/lago/vm.py
+++ b/lago/vm.py
@@ -99,6 +99,10 @@ class LocalLibvirtVMProvider(vm.VMProviderPlugin):
             with LogTask('Destroying VM %s' % self.vm.name()):
                 self.libvirt_con.lookupByName(self._libvirt_name(), ).destroy()
 
+    def restart(self):
+       self.stop()
+       self.start()
+
     def defined(self):
         dom_names = [dom.name() for dom in self.libvirt_con.listAllDomains()]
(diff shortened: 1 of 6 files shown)


mykaul:
Restart is a reboot (internal?) or stop-start of VMs? What is the use case here?

karmab:
Restart is implemented as restart for libvirt provider, but i guess
stop/start might have to be used for other virt providers.
Use case as stated in the corresponding issue is mainly ease of use of the
tool.
You also get quicker action since we parralelize on all vms.

mykaul:
@karmab  - can you get the branch up-to-date and re-test?

lago/vm.py
```diff
@@ -99,6 +99,10 @@ def stop(self):
             with LogTask('Destroying VM %s' % self.vm.name()):
                 self.libvirt_con.lookupByName(self._libvirt_name(), ).destroy()
 
+    def restart(self):
```
nvgoldin: I think it would be much easier to implement all of the above by adding a '--reboot' parameter to do_stop() , which will also start the VM. Most of the logic is already implemented in start/stop.

nvgoldin:
@karamb - closing for now, feel free to open again if you want to get this going.

PR #303, opened 2016-08-30
Removed redundant checks when getting ovirt api

gbenhaim:
wait parameter should had to be set to true
in order for any code to be executed in '_get_api',
so it is not needed.

Removed engine's service check in order to
support a situation when we can't check for the
service stauts, but https connection is available.

I think that there is a sense in trying to request
the api within a time window, instead of giving up
on first failure, So I kept the request inside
'assert_true_within_short'.

I had to remove the 'or' condition, otherwise
if '_create_api()' returns false, the condition will
be evaluated to true, which in turn will break the
loop of 'assert_true_within'. Eventually we will end
up without api object.

### DIFF
--- a/ovirtlago/virt.py
+++ b/ovirtlago/virt.py
@@ -112,26 +112,29 @@ class EngineVM(lago.vm.DefaultVM):
             insecure=True,
         )
 
-    def _get_api(self, wait):
-        if wait:
-            self.wait_for_ssh()
-            try:
-                testlib.assert_true_within_long(
-                    lambda: self.service('ovirt-engine').alive()
-                )
+    def _get_api(self):
+        try:
+            api = []
 
-                api = []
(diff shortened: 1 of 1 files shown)


mykaul:
:+1:

gbenhaim:
The lambda in my previous implementation always returned None, which implies that we will also hit the timeout (The process is stuck within on assert_true_within).

I wrote an explicit function that returns true if an api object was created successfully, otherwise keep trying until hitting the timeout.

I've tested this patch on ovirt-system-tests.

gbenhaim:
ci-test-please

gbenhaim:
ci test please

gbenhaim:
rebased

gbenhaim:
ci merge please

PR #314, opened 2016-09-11
Add serial number to disks.

mykaul:
Apparently, when using SCSI disks, you can't trust the order
UDEV will see them and enumarate them.
The only correct way is use /dev/disk/by-id/...

This adds a serial number to the disk, to make it possible.
Use /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_2 for the 1st SCSI disk, etc.


mykaul:
ci merge please

PR #312, opened 2016-09-11
Load lago.conf defaults when no config file exists

nvgoldin:
fixes https://github.com/lago-project/lago/issues/310

PR #298, opened 2016-08-24
Make lease_dir configurable

pilou-:

PR #316, opened 2016-09-12
Docs - Setup Lago in virtualenv

nvgoldin:
Note that this depends on https://github.com/lago-project/lago/pull/312 to work.

### DIFF
--- a/docs/Env_Setup.rst
+++ b/docs/Env_Setup.rst
@@ -55,7 +55,7 @@ need to run on each edition but before submitting a patch for review you should
 do it.
 
 You can run them on your local machine, but the tests themselves will install
-packages and do some changes to the os, so it's really recommmended that you
+packages and do some changes to the os, so it's really recommended that you
 use a vm, or as we do on the CI server, use mock chroots. If you don't want to
 setup mock, skip the next section.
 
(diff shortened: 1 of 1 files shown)


docs/Env_Setup.rst
```diff
@@ -108,6 +108,188 @@ with a minimal fedora 23 installation. It will leave any logs under the `logs`
...
+available on PyPI, so we have to install them from RPMs.
+
+Install virtualenv::
+
+  > sudo pip install virtualenv
+
+Add your user to ``qemu`` group and vice versa::
+
+  > sudo usermod -a -G qemu "$USER"
+  > sudo usermod -a -G "$USER" qemu
```
pilou-: An alternative to adding `qemu` to `$USER` group can be to use file access control lists (I am using `/var/lago` instead of `$HOME/lago/data`).

docs/Env_Setup.rst
```diff
@@ -108,6 +108,188 @@ with a minimal fedora 23 installation. It will leave any logs under the `logs`
...
+  > sudo mkdir -p /var/lib/lago
+  > sudo chown "$USER:$USER" /var/lib/lago
+
+Create a local ``.lago.conf`` file pointing to our new directory structure::
+
+ > cat > "$HOME/.lago.conf" << EOF
+   [lago]
+   template_store=/home/$USER/data/store
+   template_repos=/home/$USER/data/repo
+   subnet_lease_dir=/home/$USER/data/subnets
```
pilou-: Overriding `subnet_lease_dir` requires #298 :-) Once merged, the sentence above starting with `Overriding the subnet_lease_dir` could be removed.

docs/Env_Setup.rst
```diff
@@ -108,6 +108,188 @@ with a minimal fedora 23 installation. It will leave any logs under the `logs`
...
+
+Setup and activate a virtualenv, dragging the libraries already installed by
+the RPMs::
+
+  > virtualenv --system-site-packages ~/venv-lago && \
+      source "$HOME"/venv-lago/bin/activate
+
+Install Lago's Python dependencies in the newly created virtualenv::
+
+  > pip install -I enum dulwich flake8 libvirt-python lockfile \
```
pilou-: A `requirements.txt` and `build-requirements.txt` files should be created, there is already one file for the documentation (`docs/requires.txt`) and one for the tests (`test-requires.txt`).

docs/Env_Setup.rst
nvgoldin: :+1: 
noticed your PR after submitting this. I'm working on a rewrite to load all defaults in a more formalized predicted manner.

docs/Env_Setup.rst
nvgoldin: To do anything useful in the virtualenv, you would probably need the 'test-requires.txt' and not the minimalistic requirements(i.e. with flake8 and its friends), so I think having 
```pip install -r test-requires.txt ``` should suffice.

My problem with ```test-requires.txt``` is that it includes 'python-magic', and as far as I understand this is not the actual packge we are using(we use python-magic from the RPMs, which is not found on PyPi). I don't know if its there by mistake or there is some other process in ```automation``` scripts that plants the "true" package afterwards. So for now I rather list it explictly until it is resolved. (otherwise you would actually need to run ```bash pip install -r test-requires.txt && pip uninstall python-magic``` )

File: docs/Env_Setup.rst (hunk 1 above)
Comment (nvgoldin) [inline docs/Env_Setup.rst]: Can this work under ```$HOME``` directory? I think at least in virtualenv installation, we should attempt to run everything under ```$HOME```(beside global-packages of course)

File: docs/Env_Setup.rst (hunk 3 above)
Comment (pilou-) [inline docs/Env_Setup.rst]: Should we not then rename `test-requires.txt` to `requirements.txt` ? Note that `pbr` [ignores](http://docs.openstack.org/developer/pbr/#id2) `test-requires.txt`.

Comment (nvgoldin) [inline docs/Env_Setup.rst]: Thinking about my comment again, installing Lago in virtualenv can have two different use sceneario:
1. End user installing from virtualenv(and future PyPi) - then he doesn't need the testing packages.
2. Developer installing in virtualenv - then he does need the testing packages.
So I think we should still need the separation, maybe there is a common pattern to solve this, not sure.

Comment (pilou-) [inline docs/Env_Setup.rst]: A pattern is to split dependencies into separate files: for example the [repoman](https://github.com/oVirt/repoman) repository contains 4 requirements files. Other examples: [requirements for OpenStack projects](https://github.com/openstack/requirements) or [Juju-GUI](https://github.com/juju/juju-gui).

**Comment by nvgoldin** [comment]
waiting for: 
https://github.com/lago-project/lago/pull/324 - update the config file accordingly.
https://github.com/lago-project/lago/pull/327 - once this is merged, maybe differiante between installing Lago in venv for usage or foir development(test-requires.txt/requirements.txt).

File: docs/Env_Setup.rst [hunk 4]
```diff
@@ -108,6 +108,188 @@ with a minimal fedora 23 installation. It will leave any logs under the `logs`
...
+and most likely with little package-names resolving on CentOS >7.
+
+Install dependencies::
+
+  > sudo dnf install -y redhat-rpm-config libvirt libvirt-devel \
+     libguestfs-devel libguestfs-tools qemu-img qemu-kvm \
+     libvirt-daemon git gcc libffi-devel python-devel \
+     openssl-devel bats python-magic python-libguestfs
+
+Note that ``python-magic`` and ``python-libguestfs`` are currently not
```
pilou-: [file-magic](https://pypi.python.org/pypi/file-magic/0.3.0) can be used.

nvgoldin: This is great news, 
Few weeks ago I contacted the maintainer of the RPM package we use now (python-magic) and he replied they do not have packing for pip/PyPi. So I assume 'file-magic' is a different package. Were you able to test it? 
Also - If it doesn't have RPM packaging, we will still need to use the python-magic, as they might diverge with time(python-magic and file-magic).

pilou-: On Debian, lago works well with `file-magic` pip package. This package is based on the official Github [mirror](https://github.com/file/file). One author of this package is Reuben Thomas who is listed as a [contributor](https://www.openhub.net/p/file/contributors/summary) of file.
Currently, `file-magic` pip package 0.3.0 corresponds to file 5.28.

nvgoldin: If I understand correctly this means it is yet another implementation of the libmagic bindings, the ideal bindings would need to: 
1. Work with the current code(though could be that adapting it to different binding doesn't require a lot )
2. Have maintained packaging for CentOS/Fedora and Debian.
3. Available on pip.
So the only missing fact is whether ``file-magic`` has maintained packaging. For ``python-magic``, I think the only thing missing is pip. 
@gbenhaim - what you used on debian is the same ``python-magic``?

docs/Env_Setup.rst
nvgoldin: :+1:

docs/Env_Setup.rst
pilou-: I made a mistake: currently, `file-magic` pip package [0.3.0](https://pypi.python.org/pypi/file-magic/) contains https://github.com/file/file/commit/8048d19f40df4e7a2175a231fa525bc202d6b006 (a commit between 5.25 and 5.26) not 5.28. Due to a regression, this version isn't compatible with Python 2, the bug has been fixed in [file 5.26](https://github.com/file/file/commit/5560418ebc6f94e80953c8646383fdb70758fc7a). Using lago, the bug is encountered when `template_store` directory is populated.

`file` project provides [Python binding](https://github.com/file/file/blob/master/python/). `file-magic` pip package contains `magic.py` from [`file` project](https://github.com/file/file/blob/master/python/magic.py) without any modification. Debian package `python-magic` contains the binding from `file` project.

nvgoldin:
blocked by: https://github.com/lago-project/lago/issues/350
might get an intial versioning running if https://github.com/lago-project/lago/issues/350 takes too long.

nvgoldin:
closing for now until we finish moving lago-ost-plugin out and then it would simplify stuff.

PR #323, opened 2016-09-20
Add host CPU model detection

mykaul:
Added a function to detect the host CPU (that libvirt knows about).
Added a function that calculates the ovirt CPU family level.

This will later be used in ovirt-system-tests to set cluster
CPU type.

Tested with Haswell and Broadwell CPUs, on master
and hosted-engine ovirt-system-tests.

### DIFF
--- a/ovirtlago/virt.py
+++ b/ovirtlago/virt.py
@@ -33,6 +33,7 @@ class OvirtVirtEnv(lago.virt.VirtEnv):
     def __init__(self, prefix, vm_specs, net_spec):
         self._engine_vm = None
         self._host_vms = []
+        self._ovirt_cpu_family = None
         super(OvirtVirtEnv, self).__init__(prefix, vm_specs, net_spec)
 
     def _create_vm(self, vm_spec):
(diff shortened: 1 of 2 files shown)


mykaul:
For the record, this is what I've done on ovirt-system-tests to test this:
```
--- a/basic_suite_master/test-scenarios/002_bootstrap.py
+++ b/basic_suite_master/test-scenarios/002_bootstrap.py
@@ -114,12 +114,13 @@ def add_dc_quota(api):
         nt.assert_true(dc.quotas.add(quota))
 
 
-@testlib.with_ovirt_api
-def add_cluster(api):
+@testlib.with_ovirt_prefix
+def add_cluster(prefix):
+    api = prefix.virt_env.engine_vm().get_api()
     p = params.Cluster(
         name=CLUSTER_NAME,
         cpu=params.CPU(
-            id=CLUSTER_CPU_FAMILY,
+            id=prefix.virt_env.get_ovirt_cpu_family(),
         ),
         version=params.Version(
             major=DC_VER_MAJ,
```

david-caro:
I'm not sure if those tests are ok... as it depends on the machine it runs on, it might happen that the cpu family is not supported (it is supported now, and probably will be supported also in the future, so it might be ok, though if it ever happens, it cause problems as it does not seem straight forward to debug)

ovirtlago/virt.py
```diff
@@ -76,6 +77,31 @@ def engine_vm(self):
     def host_vms(self):
         return self._host_vms[:]
 
+    def get_ovirt_cpu_family(self):
+        ovirt_cpu_families = {
```
david-caro: This might be nicer as a constant, that way anyone can check the supported ones by introspection (it would be nice also to have a way to notify if the cpu is not in the supported list)

mykaul:
If the CPU is not in the supported list, then it'll use the default which is what we've always used - Conroe (which probably would have not worked on AMD, so in that sense things are improving).

This is a step towards configuring a better (more accurate CPU) for ovirt Hosted Engine VM.
If there is a newer CPU, we'll need to handle it. (I can think of a host with a newer CPU running an older oVirt version, for example). It should be fairly easy to fix in a single function most likely.

It improves the testing coverage, but giving us more opportunity to test more variants.

mykaul:
ci merge please

Issue #58, opened 2016-01-27
Add configuration file for all the options


david-caro:
Moved from https://bugzilla.redhat.com/show_bug.cgi?id=1277656

Right now there's no way to configure lago system options like repo caches, template caches and such, adding a way to specify a conf file where to override any of those hardcoded values will add a lot of ease on adapting it to unexpected systems

Issue #310, opened 2016-09-07
lago init does not fallback to defaults when there is no lago.conf


nvgoldin:
The defaults are defined in [lago/configs.py]
(https://github.com/lago-project/lago/blob/27856ea92da94b5816c4c84acd02247fb999af54/lago/config.py#L27-L41)
```python
DEFAULTS = {
    'default_root_password': '123456',
    'default_ssh_password': '123456',
.....
    'libvirt_password': '',
}
```

But they aren't loaded, the reason is that in [configs.py: def _get_from_files](https://github.com/lago-project/lago/blob/27856ea92da94b5816c4c84acd02247fb999af54/lago/config.py#L52-L58) method:
```python
def _get_from_files(paths, key):
    config = ConfigParser.ConfigParser(defaults=DEFAULTS)
    config.read(paths)
    try:
        return config.get('lago', key)
    except ConfigParser.Error:
raise KeyError(key)
```
```ConfigParser.ConfigParser``` method loads ```DEFAULTS``` into the special 'default' section, and not to the 'lago' section. If the file already exists this might work, but when it doesn't ```config.get('lago', key)``` will always fail(there is no ``lago`` section).

nvgoldin:
closed in https://github.com/lago-project/lago/pull/324

Issue #137, opened 2016-02-23
Allow to configure the libguestfs cache and tmp dirs


david-caro:
From the #94:

> That's a good idea - I use it without any issues for some time now. Two more enhancements:
> export LIBGUESTFS_CACHEDIR=/home/zram
> export LIBGUESTFS_TMPDIR=/home/zram

nvgoldin:
Testing I have done with those parameters didn't yield major improvements. I  think bootstrap stage is pretty fast. Also, as we use scp in collect now, bootstrap/customize is the only gain.

Second, you can always run ``export LIBGUESTFS...`` prior to running Lago and it will use them.

Issue #60, opened 2016-01-27
template_store and template_repos should be part of enviroment


david-caro:
Moved from https://bugzilla.redhat.com/show_bug.cgi?id=1272034

Those two dirs tend to be very large in size and files may collide between projects
Not everybody has free 20GB on their / dir
They should reside in the working directory.

ifireball:
I'd rather see a per-user shared cache in the XDG standard location then
one per environment, as its quite likely that multiple environments used by
the same user will share images.

On 27 January 2016 at 21:39, David Caro <notifications@github.com> wrote:

> Moved from https://bugzillaredhatcom/show_bugcgi?id=1272034
>
> Those two dirs tend to be very large in size and files may collide between
> projects
> Not everybody has free 20GB on their / dir
> They should reside in the working directory
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/lago-project/lago/issues/60>.
>



-- 
*Barak Korren*
bkorren@redhat.com
RHEV-CI Team

nvgoldin:
So the idea is to store images in
~/.cache/lago/store
~/.cache/lago/repos

Which will be auto-created if they do not exist. 
Only question left ATM is subnets - we need a world writeable/readable, as this is a system-wide resource, even if you install it via pip.

Also this means that once we make this change - images will be downloaded again for most users..(unless they change the location in lago.conf)

mykaul:
On Jun 19, 2017 6:28 PM, "Nadav Goldin" <notifications@github.com> wrote:

So the idea is to store images in
~/.cache/lago/store
~/.cache/lago/repos


Isn't there some /var/cache that can be used?

Which will be auto-created if they do not exist.
Only question left ATM is subnets - we need a world writeable/readable, as
this is a system-wide resource, even if you install it via pip.

Also this means that once we make this change - images will be downloaded
again for most users..(unless they change the location in lago.conf)

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/lago-project/lago/issues/60#issuecomment-309475786>,
or mute
the thread

nvgoldin:
> Isn't there some /var/cache that can be used?

/var/cache will not be per-user, so if you use 'pip install' you'd still have to create those directories with appropriate permissions(for all users), so that is the same as /var/lib I think, though will need to check

You can still always configure it to a shared location in lago.conf, the question is what would be the default.

PR #324, opened 2016-09-20
Load configurations from the CLI plugins

nvgoldin:
This PR attempts to fix few problems with how Lago loads configurations:

1. Defaults are not unified - Some are defined in /etc/lago.d/lago.conf, some using the plugin cli system and some are in config.py DEFAULTS dict.
2. Environment variables support is not full - you can't pass environment variables to a sub-parser without having them in a global scope.
3. lago.conf is not self-contained: The RPM puts the hard coded config file at /etc/lago.d/lago.conf, which is a problem when you want to install Lago in user-space.
5. There is no standard configuration load order(see https://github.com/lago-project/lago/issues/60)

The approach taken to fix all of the above, is to use the already-existing plugin system, and loading all defaults from the CLI plugin system(using ```lago.plugins.cli.cli_plugin_add_argument``` decorator). This has few advantages:

1. Single point for defining  defaults, ```default=...``` in the decorator.
2. Auto-generating the configuration files easily.
3. It makes everything configurable via cli arguments, enviornment variables and INI configs. 
4. If something is not configurable, it goes somewhere else(hard-coded, constants.py, etc)

Implementation details
-----------------------------
1. All configuration files are loaded according to slightly modified XDG standards:
  * Load /etc/lago/lago.conf if exists
  * Load files from XDG standard(if not set otherwise by the user defaults to: ```/etc/xdg/lago/lago.conf, $XDG_HOME_PATH/.config/lago/lago.conf```)
  * ENV variables
  * CLI passed arguments.
  * The later always takes precedence. 

2. All variables are defined using the CLI plugin system, this means:
  * Each ``verb``(init, start, etc) declares its own local scope variables, this will go automatically to the config INI files under ```[verb]``` section.
  * Global variables are in the parser root argument.

3. A new verb is added ``lago generate-config`` which dumps the current loaded configuration to stdout, this allows:
  * When generating the RPMs create a default commented-out /etc/lago/lago.conf, dynamically from the declared CLI arguments. That way we no longer need to maintain manually lago.conf
  * When a user installs Lago, for quick break in he can run:
```lago generate-config > ~/etc/.config/lago/lago.conf```  and edit it to his needs.

4. All sub-parsers arguments can be passed using environment variables:
```
export LAGO_{global parameter name}="global_param_value"
export LAGO__{verb}__{verb_specific_parameter}="a_verb_param_value"
```
Example of using environment variables:
```
> # global config
> lago generate-config | grep ssh_tries
ssh_tries = 100
> export LAGO_SSH_TRIES=30
> lago generate-config | grep ssh_tries
ssh_tries = 30
> # verb config
> export LAGO__INIT__TEMPLATE_STORE=/tmp/lago/store
> lago generate-config | grep template_store
template_store = /tmp/lago/store
> lago init --help | grep -B2 'tmp'
  --template-store TEMPLATE_STORE
                        Location to store templates at (default:
                        /tmp/lago/store)
```

New verb ``generate-config`` example, which is also the new ``lago.conf`` format: 
```bash
> lago generate-config
# Lago configuration file, generated: Wed Sep 21 16:04:19 2016
[lago]
loglevel = info
# log level to use
logdepth = 3
# how many task levels to show
out_format = default
prefix_name = current
# name of the prefix to use.
ssh_user = root
# user for ssh provider.
ssh_password = 123456
# password for ssh provider.
ssh_tries = 100
# number of ssh time outs to wait before failing.
ssh_timeout = 10
# seconds to wait before marking ssh connection as failed.
libvirt_url = qemu:///system
# libvirt uri, currently only system is supported.
default_vm_type = default
# default vm type
default_vm_provider = local-libvirt
# default vm provider
default_root_password = 123456
# default root password
lease_dir = /var/lib/lago/subnets
# path to store created subnets configurations
reposync_dir = /var/lib/lago/reposync
# reposync dir if used
reposync_config = config.repo
# reposync config
ignore_warnings = False

[init]
skip_bootstrap = False
# if passed, will skip bootstrapping the images, useful if you are using
# templates and you already know they will have the correct root pass
# for example
set_current = False
# if passed, it will set the newly created prefix as the current one
template_repos = /var/lib/lago/repos
# location to store repos
template_store = /var/lib/lago/store
# location to store templates at
template_repo_path = http://templates.ovirt.org/repo/repo.metadata
# repo file describing the templates

[snapshot]
list_only = False
# list current available snapshots

[destroy]
yes = False
# don't ask for confirmation, assume yes
all_prefixes = False
# destroy all the prefixes in the workdir

[generate-config]
defaults_only = False
# ignore cli parameters and print loaded configs only.
```

Future work could include having other plugins configuration auto-generated too(VM, Service, Provider, etc), so eventually we will have  sections such as ```[libvirt], [ssh], etc``` in the INI file.

closes: https://github.com/lago-project/lago/issues/58, https://github.com/lago-project/lago/issues/60, https://github.com/lago-project/lago/issues/310
affects: https://github.com/lago-project/lago/issues/137, https://github.com/lago-project/lago/issues/47, https://github.com/lago-project/lago/pull/316


david-caro:
Amazing, I love it!! +100!

lago/cmd.py
```diff
@@ -83,6 +83,13 @@
 @lago.plugins.cli.cli_plugin_add_argument(
     '--template-store',
     help='Location to store templates at',
+    default='/var/lib/lago/store',
```
david-caro: is there a way to 'interpolate' the values? I say because it might be nicer to use something like '{base_dir}/store' so you don't have to repeat everywhere and it also makes it easier to change. A quick solution might be to just add the base path as another param and make these others relative.

lago/cmd.py
```diff
@@ -154,24 +162,17 @@ def do_init(
                 template_repo_path
             )
         else:
-            try:
-                repo_name = (
-                    template_repo_name
-                    or lago.config.get('template_default_repo')
```
david-caro: Careful! this might break old configs (as it's not using the template_default_repo value anymore)

lago/cmd.py
```diff
@@ -578,22 +579,34 @@ def do_deploy(prefix, **kwargs):
...
 
+@lago.plugins.cli.cli_plugin(help="Dump configuration file")
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--defaults_only',
+    help='Ignore CLI parameters and print loaded configs only.',
+    action='store_true',
+    default=False
+)
+def do_generate(defaults_only, **kwargs):
+    print config.get_ini(defaults_only=defaults_only)
```
david-caro: this is very useful!

lago/cmd.py
```diff
@@ -623,23 +636,110 @@ def create_parser(cli_plugins, out_plugins):
...
+    )
+
+    parser.add_argument('--ignore-warnings', action='store_true')
+    parser.set_defaults(**config.get_section('lago', {}))
     verbs_parser = parser.add_subparsers(dest='verb', metavar='VERB')
     for cli_plugin_name, cli_plugin in cli_plugins.items():
         plugin_parser = verbs_parser.add_parser(
-            cli_plugin_name, **cli_plugin.init_args
+            name=cli_plugin_name,
+            formatter_class=argparse.ArgumentDefaultsHelpFormatter,
```
david-caro: This is the bit that fixes the 'help command is not helpful' issue?

lago/plugins/vm.py
```diff
@@ -36,8 +36,8 @@
 
 from scp import SCPClient
 
-from .. import (config, utils, log_utils, plugins, ssh, )
-
+from .. import (utils, log_utils, plugins, ssh, )
+from lago.config import config
```
david-caro: missing newline

ovirtlago/virt.py
```diff
@@ -50,7 +51,7 @@ def _create_vm(self, vm_spec):
         else:
             provider_name = vm_spec.get(
                 'vm-type',
-                lago.config.get('default_vm_provider', 'default'),
+                lago_config.get('default_vm_provider', 'default'),
```
david-caro: Is there any advantage to this? (just curious)

tests/unit/test_config.py
```diff
@@ -1,125 +1,97 @@
...
+        )
+        res = cfg.load()
+
+        assert mocked_configs_path.call_count == 1
+        assert mocked_open.call_args_list == [
+            call('file1', 'r'), call(
+                'file2', 'r'
+            )
+        ]
+        assert res == {'section': {'var1': 'file2'}}
```
david-caro: Maybe missing the command line override everything check

lago/config.py
```diff
@@ -14,88 +14,216 @@
...
+
+    """
+    env_lago = defaultdict(dict)
+    regex = r'((?P<nosect>^{0}_(?!_))|(?P<sect>^{0}__))(?P<name>.*$)'.format(
+        re.escape(root_section.upper())
+    )
+    decider = re.compile(regex)
+    for key, value in os.environ.iteritems():
+        match = decider.match(key)
+        if match:
```
david-caro: As the inner block is big, I'd suggest using the negative check to skip the loop instead (and dedent it):

```python
if not match:
    continue
<inner block>
```

lago/config.py
```diff
@@ -14,88 +14,216 @@
...
+            if not match.group('name') or not value:
```
david-caro: Shouldn't this be `if match.group('name') and not value`? (that is, the var is defined but empty)

lago/config.py
```diff
@@ -14,88 +14,216 @@
...
+                warn(
+                    'empty environment variable definition:'
+                    '{0}, ignoring.'.format(key)
+                )
+            elif match.group('nosect'):
+                env_lago[root_section][match.group('name').lower()] = value
```
david-caro: maybe assign the groups to variables to make more readable this part of the code, something like:

```python
_, name, section, global = [word.lower() for word in match.groups()]
```

It feels a bit strange too that you have the sect group in the regexp but then use split to extract it from the name, maybe you should try to get it already from the regexp (if you haven't tried, if it gets too complicated it's not worth it imo)

lago/config.py
```diff
@@ -14,88 +14,216 @@
...
+        self._parser = parser
+        ini_str = argparse_to_ini(parser)
+        configp = configparser.ConfigParser(allow_no_value=True)
+        configp.read_dict(self._config)
+        configp.read_string(ini_str)
+        self._config.update(
+            {
+                s: dict(configp.items(s))
+                for s in configp.sections()
+            }
```
david-caro: Just to make sure, this is not going to run on python<2.6 right? (I really hope not, but if it is, this dict comprehension is not supported)

lago/config.py
```diff
@@ -14,88 +14,216 @@
...
+        return self._config.get(*args)
+
+    def get_ini(self, defaults_only=False):
+        """Return the config dictionary in INI format
+        Args:
+            defaults_only (bool): if set, will ignore arguments set by the CLI.
+
+        Returns:
+            str: string of the config file in INI format
+        """
```
david-caro: Feels strange that sometimes there is a newline after the docstring and sometimes it's not (not really an issue for me though, just noticed)

lago/config.py
```diff
@@ -14,88 +14,216 @@
...
+                return out_ini.getvalue()
+
+    def __repr__(self):
+        return self._config.__repr__()
+
+    def __str__(self):
+        return self._config.__str__()
+
+
+config = ConfigLoad()
```
david-caro: Is this really necessary? Can't it be done on demand instead? (to avoid instantiating it on import, and running it instead on demand)

lago/cmd.py
nvgoldin: I see no evidence of 'template_default_repo' parameter in our code-base, of course could be that users used it. You know if we actually used it anywhere?

As a side note, currently ```generate-config```  filters CLI arguments that don't have default values, such as template_repo_name(you can still add it there or in ENV), but maybe a better approach would be to leave them commented out, so their existence won't be forgotten.

lago/cmd.py
nvgoldin: Nope, reopend https://github.com/lago-project/lago/issues/98, I misread the title, ``lago shell --help`` is still broken ;/


``ArgumentDefaultsHelpFormatter`` - prints loaded defaults in help commands.

ovirtlago/virt.py
nvgoldin: later there is another 'config' variable used, didn't want to mask it.
```python
def engine_setup(self, config=None):
....

```

lago/config.py
nvgoldin: it checks both - either  ``LAGO_/LAGO__`` was declared or a variable with no value was declared:
```bash
> export LAGO_=value
> export LAGO__=value
> export LAGO_EMPTY_GLOBAL=
> export LAGO__VERB__EMPTY_VAR
> lago help
.../lago/lago/config.py:69: UserWarning: empty environment variable definition:LAGO_, ignoring.
  ('empty environment variable definition:''{0}, ignoring.').format(key)
.../lago/lago/config.py:69: UserWarning: empty environment variable definition:LAGO__VERB__EMPTY_VAR, ignoring.
  ('empty environment variable definition:''{0}, ignoring.').format(key)
.../lago/lago/config.py:69: UserWarning: empty environment variable definition:LAGO__, ignoring.
  ('empty environment variable definition:''{0}, ignoring.').format(key)
.../lago/lago/config.py:69: UserWarning: empty environment variable definition:LAGO_EMPTY_GLOBAL, ignoring.
  ('empty environment variable definition:''{0}, ignoring.').format(key)

```

lago/config.py
nvgoldin: Good point, would probably fail(don't have 2.6 env running here to test), we don't support it right?(in the docs I see 2.7)

lago/config.py
nvgoldin: I pondered on this issue too. I can't say I'm happy with the result, but couldn't find any better pattern, some of the alternatives:
1. Drop ``ConfigLoad`` class and implement everything in as module-functions. This has some overhead of needing to check in each function if the dict was already populated, and overall would make the code a bit more clumsy(no ```__init__```), and less modular. 
2. Implement it as a singleton(as it is now, sort of), I looked on different python approaches to this(metaclass, class decorators, etc), but all these approaches would make the API look like:
```python
# get the singleton class
config = get_config() / config.Config() 
# consume it
config.get('..')
# or more clumsy and less readable
config.Config().get('...')
```
this would add some reduandant lines in the entire code base.

The last approach was to 'create' the class at module level - this has the overhead of 'weird' import:
```from .config import config```
but lets everyone use the same API as before(as if config is a dict).

would be happy to hear if there is a better way :)

lago/cmd.py
nvgoldin: looks like [argparse supports](https://docs.python.org/3/library/argparse.html#help) interpolation only for the ``help`` keyword, the INI part should be easier as ``configparser`` supports interpolation. Either way to make generic support would involve some in-house solution.
If we change it to base_dir, wouldn't that be too restrective on the users? (having store, repos, subnets directories in one place, while one might want to move store or repos to a different device with more space for instance )

gbenhaim:
Great job !

lago/cmd.py
david-caro: About the 'template_default_repo', I'm not aware, but well, you might want to ask around (not sure if the foreman guys are still using lago or where it's being used right now). In any case, if compatibility is broken, I'd expect a major version bump.

About the 'generate-config' yes, that would be nicer, though I'd leave it for a later pr, as this one is big enough and that is not required right now, up to you though.

lago/config.py
david-caro: Hmmm, though I think that there might be configuration values that might be empty no? Maybe we should do the content checking in another way... (not blocking though, might be a future improvement)

lago/config.py
david-caro: Iirc we were using it somewhere on upstream but it was about to be deprecated, not sure of the current status, maybe you will want to rise the issue on the weekly meeting.

lago/config.py
david-caro: Hmmm, I think that it would be possible to make the singleton instance lazy, in the sense that it does not get instantiated until someone tries to access any of it's methods. Maybe just override the __getattribute__ method and there load the singleton and proxy everything.... that might make the class code a bit ugly though.

Another idea is to implement at the module top level the 'get' functions, and then on those do the lazy loading of the singleton, though you will lose all the 'like a dictionary' features (hmm, am I right? What happens if you define the __getitem__ at the module level?... a longshot though xd, for sure it's treated differently as there's no self, aka. it's not a bounded method to the module).

In any case, I don't think this is a blocker either, we can do the refactor later if needed if you don't see a clear advantage to any.

ovirtlago/virt.py
david-caro: With the old `lago.config` it was not masked, as the other is namespaced behind the `lago` module, but :+1: for the initiative (no need to revert this, it's ok as it is too).

lago/config.py
nvgoldin: got it working with group reference in the regexp - https://github.com/lago-project/lago/pull/324/commits/afae55c5271b54d0cf3858ddefc5ec2aa949c4bb
the regexp is a bit more complicated, but still readable I think.

tests/unit/test_config.py
nvgoldin: fixed, added some more tests too.

lago/config.py
nvgoldin: thanks, fixed.

nvgoldin:
@david-caro thanks for the review :)
This is ready on my side, last commit updates the .spec file to auto-generate the config file.

I want to get this going for next version if anyone else wants to review.

nvgoldin:
squashed commits and rebased

nvgoldin:
ci merge please

nvgoldin:
ci merge please

nvgoldin:
ci merge please

PR #325, opened 2016-09-22
Added the path to qemu-kvm binary in Ubuntu

gbenhaim:
Needed for running Lago on Ubuntu.

### DIFF
--- a/lago/vm.py
+++ b/lago/vm.py
@@ -24,6 +24,7 @@ import logging
 import lxml
 import os
 import pwd
+import platform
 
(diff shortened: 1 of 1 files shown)


lago/vm.py
```diff
@@ -229,11 +230,16 @@ def _libvirt_xml(self):
         qemu_kvm_path = [
             path
             for path in [
+                '/usr/bin/qemu-system-{arch}'.format(
+                    arch=platform.processor()
+                ),
```
pilou-: See the related PR #296 and the question on architecture.

Issue #309, opened 2016-09-06
Packaging Lago for non rpm systems


gbenhaim:
We want to create .deb and python packages for lago.

Initial investigation can be found in the following link: https://github.com/lago-project/lago-demo/pull/6

kivio:
If you think about sharing Lago on PyPi there is few thinks for discussion:
1. what sens have concentrating on system specific packages -  in this case most important is python package, others are just addition
2. If we want to add support for containers and other virtualization methods then dependency for libvirt must be fast move to be plugin lazy loaded on use, it dips will be installed just if user want to use it
3. Is important to remove ovirt specific things (this is not place for that), there should be in ovirt test repositories. If is required to have ovirt here maybe it can be plugin as other virtualization method
4. Is needed to make possible to install develop version of Lago without configuration, currently it give exception because installation isn't found

pilou-:
In order to run the oVirt test suite on Debian in a Python 2.7 virtual environment created with `--system-site-packages`, the following packages are required (at least):
* `libvirt-daemon-system`, `qemu-system-x86`, `python-guestfs`, `libguestfs-tools`, `libxslt1-dev`, `python-magic`, `yum-utils`, `koji-client`, `python-rpm`, `createrepo` Debian packages
* pip packages available on PyPI: `ovirt-engine-sdk-python` and the one listed in requirements.txt files

* `repoman` pip package (not the one available on PyPI):
    ```
    git clone https://github.com/oVirt/repoman
    pip install -r build-requirements.txt
    pip install -e .
    ```
* Fix for QEMU path: #296 or #325

nvgoldin:
Currently blocked by: https://github.com/lago-project/lago/issues/350

hagzag:
What is the ETA on support for non RedHat OS's e.g Ubuntu / Debian - I am now at a decision point of replacing Vagrant with a more flexible system ...

eedri:
I think we are very close to have a formal Python packaging, see #363 and #350. 
Once we'll have Python packaging in place, adding other OS packaging like Debian will be much easier.
@gbenhaim @nvgoldin  maybe have more detailed info on it.

mykaul:
@hagzag - why are you considering replacing vagrant? Would be good to know the reasons.

gbenhaim:
Lago is available in PyPy.

PR #327, opened 2016-09-25
Added requirements.txt

gbenhaim:
This file contains all the deps
that can be downloaded from PyPi.

Other deps should be downloaded with
the platform's native package manager.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,9 @@
+lxml
+lockfile
+paramiko
+xmltodict
+scp
+setuptools
+stevedore
+PyYAML
+enum


gbenhaim:
Part of #309

nvgoldin:
:+1: 
I guess libvirt-python is not working properly on ubuntu from PIP?

requirements.txt
```diff
@@ -0,0 +1,10 @@
+lxml
+lockfile
+paramiko
+pbr
```
pilou-: [`dulwich`](https://github.com/lago-project/lago/blob/adding_requirements.txt/scripts/version_manager.py#L10) (used in  `scripts/version_manager.py` which is called by [`setup.py`](https://github.com/lago-project/lago/blob/adding_requirements.txt/setup.py#L37)), [`file-magic`](https://github.com/lago-project/lago/blob/adding_requirements.txt/lago/templates.py#L26) and `libvirt-python` are missing.

nvgoldin:
dulwich - is used only for building/testing as far as I know.
file-magic - is taken from the RPM(python-magic). What that is available on PyPi is a different package.

gbenhaim:
@nvgoldin 
You are getting it for free when installing libvirt-bin :)

pilou-:
`pbr` is used only for building, but this package is listed: either both `pbr` and `dulwich` should be listed or both should not.

[file-magic ](https://pypi.python.org/pypi/file-magic/0.3.0) pip package contains [magic.py](https://github.com/file/file/blob/master/python/magic.py) released in file 5.26.

gbenhaim:
Removed pbr

gbenhaim:
ci merge please

PR #331, opened 2016-09-27
Let pbr handle version

pilou-:
`pbr` is already a build dependency and is able to compute package version: helper method `get_current_version` defined in `scripts/version_manager.py` isn't required.


david-caro:
Note that pbr does not handle the versioning the same way and does not generate the changelog required for the rpm packaging (I'm also working on adding a release notes option to the version script, that I'm using on another poject too).

I don't think there's any advantage yet to switch to pure pbr right now, the version script code has not been a burden so far and we can't get rid of it with this change.

pilou-:
`pbr` is able to [generate](http://docs.openstack.org/developer/pbr/api/pbr.html?highlight=generate#pbr.git.write_git_changelog) the changelog.

Advantages would be:
* use well known [semantic versioning 3.0](http://docs.openstack.org/developer/pbr/semver.html) / avoid to use a custom version number management
* generate ChangeLog / AUTHORS from git log
* remove a build-dependency (`dulwich`)

@david-caro should we update this pull-request in order to use `pbr` for the changelog creation ?

david-caro:
I'm looking into the current pbr status and way of generating the versions, for now I can say that we are already (and more strictly) using semantic versioning (2 and 3), and that we are also generating the changelog from the git log, so the only issue is the extra build dependency (that in pbr is hidden, as it requires git underneath instead).

But if we can get rid of the code without giving anything up (release notes, current versioning, rpm suitable changelog, etc...) would be great.

We started using pbr at first, but the versioning was not nice (it required signed tags to actually detect that it's a release, and to do the actual semantic versioning, while we want to be able to release any commit, tagged or not as the tagging process is not atomic), and to have an rpm suitable changelog so the version script was born.

nvgoldin:
Hi @pilou-, you still want to get this going? 
I think if so - the best approach would be to actually try and do a 'phony' release without the custom function to prove we can get it right without. Take into account you'll need to have a look at the ``automation`` directory, I also have somewhere a google doc describing the steps to do a release, if you want it ping me on irc (_ng freenode/oftc), probably should put it on the docs section.

Personally I rather use pbr  instead of the custom scripts, if it does give us what we need easily.

david-caro:
Fyi I've bundled the scripts together, added some tests and improved the usability here https://github.com/david-caro/python-autosemver, the only thing missing to use on lago is the pr (started working on it) and packaging it as rpm so we can install for the tests and such, that would completely replace pdb with a way more simple module.

nvgoldin:
@dcaro - sounds good!
that would mean completely removing pbr?

david-caro:
Yep, now it also generates the authirs (just as pbr, dummy), generates changelog for rpm and non-rpm format, and some sort of release notes (need improving) where you have the list of features, breaking changes and bugs since a commit.

We can move it to ovirt org if you want to use it and not depend on some 'personal' project, maybe use jenkins and standard-ci for the rpm building and travis for pip releasing (or try to get jenkins to deploy on travis).

nvgoldin:
@david-caro - any news on that?
I'm closing this specific PR as it has no activity(we can continue on a new issue..)

PR #332, opened 2016-09-28
Moving over get_prefixed_name() to ovirtlago/testlib

mykaul:
This is more useful than having it copied around in multiple tests.
Just a handy function.

After it'll be included in a released Lago, will convert
ovirt-system-tests to use it.

### DIFF
--- a/ovirtlago/testlib.py
+++ b/ovirtlago/testlib.py
@@ -46,6 +46,13 @@ def get_test_prefix():
     return _test_prefix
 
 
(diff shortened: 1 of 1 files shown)


mykaul:
No, I'd like to use '-' since this will ensure hosts will not have an underscore in their name (a bug I'm working on fixing in ovirt-system-tests - which depends on this).

gbenhaim:
+1

mykaul:
ci merge please

PR #333, opened 2016-09-28
Change to repoman options

gbenhaim:
This change allows repoman to collect
multiple versions of the same package into
the internal repo of the prefix.

Currently packages are being taken to the
internal repo in the following order:
1. sources that were specified using '--custom-source'.
2. Packages from the repos that were specified in the
   reposync-config file (passed to lago with --reposync-yum-config).
   Note: currently, the order in the config matters, for example:
   only package 'x.n' from the first repo will be added to the internal
   repo, even if package 'x.n+1' is in the second repo.

This situation causes unwanted behaviour when there are multiple
versions of the same package in different repos, and this change
is meant to fix it.

Signed-off-by: gbenhaim galbh2@gmail.com

### DIFF
--- a/ovirtlago/__init__.py
+++ b/ovirtlago/__init__.py
@@ -226,14 +226,14 @@ class OvirtPrefix(Prefix):
 
             rpm_dirs.extend(
                 [
-                    os.path.join(folder, dist) + ':only-missing'
+                    os.path.join(folder, dist)
                     for folder in project_roots if os.path.exists(folder)
                 ]
             )
 
             rpm_dirs.extend(
                 [
-                    os.path.join(repos_path, name) + ':only-missing'
+                    os.path.join(repos_path, name)
                     for name in repo_names if name.endswith(dist)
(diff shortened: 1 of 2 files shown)


gbenhaim:
@david-caro 
Thoughts ?

david-caro:
This was meant to be a way to specify a custom repo (for example, the factory repos built for ovirt) as the last parameter, so you just pull in the rpms that are missing from there. Allowing you to test custom-built rpms that have a lower version than the rpms in the 'base' repo, something that we already wanted numerous times.

So the process usually was:
* Create all the custom rpms that you want to test (for example, vdsm).
* Select the base repo with all the other required 'stable' rpms live, like the ovirt snapshot.
* Create a mixed repo (the internal repo) with the custom rpms + any missing rpm from the base repo.
* Run the tests with that.

If that changed, this is not needed anymore, but as this was an expected behavior this change might require a major version bump.

PR #338, opened 2016-09-30
Prep for oVirt v4 API

mykaul:
Specifically refer to v3 API for now.
Next item would be to actually add v4 API.

### DIFF
--- a/ovirtlago/testlib.py
+++ b/ovirtlago/testlib.py
@@ -70,6 +70,19 @@ def with_ovirt_api(func):
     return wrapper
 
 
+def with_ovirt_api4(func):
+    @functools.wraps(func)
+    @with_ovirt_prefix
+    def wrapper(prefix, *args, **kwargs):
+        return func(
+            prefix.virt_env.engine_vm().get_api(api_ver=4),
+            *args,
+            **kwargs
+        )
+
+    return wrapper
+
+
 def continue_on_failure(func):
     func.continue_on_failure = True
     return func
@@ -132,8 +145,9 @@ def test_sequence_gen(test_list):
                     failure_occured[0] = True
                 raise
 
-        wrapped_test.description = test.__name__
-        yield wrapped_test
+        if test is not None:
+            wrapped_test.description = test.__name__
+            yield wrapped_test
 
 
 class LogCollectorPlugin(nose.plugins.Plugin):
--- a/ovirtlago/virt.py
+++ b/ovirtlago/virt.py
@@ -25,6 +25,11 @@ import lago.config
 import lago.vm
 import ovirtsdk.api
 from ovirtsdk.infrastructure.errors import (RequestError, ConnectionError)
+try:
+    import ovirtsdk4 as sdk4
+    API_V4 = True
+except ImportError:
+    API_V4 = False
 
 from . import (constants, testlib, )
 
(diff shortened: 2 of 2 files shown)


david-caro:
Hmm... maybe better use a generic 'create_api' and then use the version as a parameter instead of adding one property for each api version no?

mykaul:
The reason I preferred NOT to do it, is that I think we'll have a long time (forever?) where we'd like to actually use, in parallel, both APIs. So I can see a test suite that is asking for both APIs. Which is why I'd like to have a get_api_v3() and get_api_v4() available. Of course, I can have a create_api(version) and both will call it if you prefer.

mykaul:
@david-caro - switched to your idea. Hope it looks better now.

ovirtlago/virt.py
```diff
@@ -118,34 +118,35 @@ def wait_for_ssh(self):
...
-            insecure=True,
-        )
+        if api_ver == 3:
+            return ovirtsdk.api.API(
+                url=url,
+                username=constants.ENGINE_USER,
+                password=str(self.metadata['ovirt-engine-password']),
+                validate_cert_chain=False,
+                insecure=True,
+            )
```
david-caro: I'd add an else that throws something like 'UsupportedApiVersion' exception or similar too (to help debugging)

ovirtlago/virt.py
```diff
@@ -118,34 +118,35 @@ def wait_for_ssh(self):
...
 
-    def _get_api(self):
+    def _get_api_v3(self):
```
david-caro: This does not need to be renamed at all no? just add the api_ver param and pass it to the _create_api, and then this can be reused for any version.

ovirtlago/virt.py
```diff
@@ -156,12 +157,15 @@ def get():
...
     def get_api(self):
-        if self._api is None or not self._api.test():
-            self._api = self._get_api()
-        return self._api
+        return self.get_api_v3()
+
+    def get_api_v3(self):
+        if self._api_v3 is None or not self._api_v3.test():
+            self._api_v3 = self._get_api_v3()
+        return self._api_v3
```
gbenhaim: Good idea.

mykaul: I sincerely hope we are not going to have more major versions any time soon, so I think this is premature. I will add the exception though - makes sense.
I'll also actually add v4 API and test to see this is all not breaking...

ovirtlago/virt.py
mykaul: Done.

mykaul:
I think this round is a cleaner approach. Tested it with a modified ovirt-system-test that calls a v4 API.

ovirtlago/testlib.py
```diff
@@ -70,6 +70,19 @@ def wrapper(prefix, *args, **kwargs):
     return wrapper
 
 
+def with_ovirt_api4(func):
```
david-caro: nice

ovirtlago/virt.py
```diff
@@ -25,6 +25,11 @@
 import lago.vm
 import ovirtsdk.api
 from ovirtsdk.infrastructure.errors import (RequestError, ConnectionError)
+try:
+    import ovirtsdk4 as sdk4
+    API_V4 = True
+except:
```
david-caro: You might want to catch only the ImportError exceptions here.

ovirtlago/virt.py
```diff
@@ -118,34 +123,51 @@ def wait_for_ssh(self):
...
+        raise RuntimeError('Unknown API requested: %s' % api_ver)
 
-    def _get_api(self):
+    def _get_api(self, api_ver):
         try:
-            api = []
+            api_v3 = []
+            api_v4 = []
 
             def get():
```
david-caro: I don't like the usage of the closure variables here to store the return values instead of using regular explicit return, but that's a different change I guess.

gbenhaim: We are using the closure because ```testlib.assert_true_within``` isn't returning the return value of the function that it runs.

I think that we should keep ```testlib.assert_true_within``` the same it is now, as it allow us to separate between the return value of the inner function and the result of the timeout assertion (for example a case where the assertion failed but we still want to get the return value of the inner function).

david-caro: I don't agree, right now `testlib.assert_true_within` does not return anything right now, it uses exceptions to handle timeouts (I agree with that), but it would be very useful if the return value would be the one of the function it executes, allowing non-implicit tricks to get results from it (like using the closure variables).

gbenhaim:
ci test please

gbenhaim:
ci merge please

gbenhaim:
ci merge please

PR #340, opened 2016-10-06
Add possibility of using custom VM XML

mpolednik:
Lago only has a single libvirt XML in it's code base. This approach is limiting for testing of more complex features such as device assignment. We therefore allow users to specify custom libvirt XML while keeping lago-specific replacements available.

### DIFF
--- a/lago/prefix.py
+++ b/lago/prefix.py
@@ -1182,6 +1182,14 @@ class Prefix(object):
             for host_name, host_spec in domains.iteritems():
                 host_metadata = host_spec.get('metadata', {})
                 deploy_scripts = self._get_scripts(host_metadata)
+                if host_metadata.get('custom-xml', None):
+                    host_metadata['custom-xml'] = os.path.expanduser(
+                        os.path.expandvars(
+                            self._copy_deploy_scripts(
+                                [host_metadata['custom-xml']]
+                            )[0]
+                        )
+                    )
                 new_scripts = self._copy_deploy_scripts(deploy_scripts)
                 self._set_scripts(
                     host_metadata=host_metadata,
--- a/lago/prefix.py
+++ b/lago/prefix.py
@@ -1182,7 +1182,7 @@ class Prefix(object):
             for host_name, host_spec in domains.iteritems():
                 host_metadata = host_spec.get('metadata', {})
                 deploy_scripts = self._get_scripts(host_metadata)
-                new_scripts = self._copy_delpoy_scripts(deploy_scripts)
+                new_scripts = self._copy_deploy_scripts(deploy_scripts)
                 self._set_scripts(
                     host_metadata=host_metadata,
                     scripts=new_scripts,
@@ -1190,7 +1190,7 @@ class Prefix(object):
 
         return domains
 
-    def _copy_delpoy_scripts(self, scripts):
+    def _copy_deploy_scripts(self, scripts):
         """
         Copy the given deploy scripts to the scripts dir in the prefix
 
--- a/lago/vm.py
+++ b/lago/vm.py
@@ -255,8 +255,16 @@ class LocalLibvirtVMProvider(vm.VMProviderPlugin):
             dom_raw_xml = dom_raw_xml.replace(key, str(val), 1)
 
         dom_xml = lxml.etree.fromstring(dom_raw_xml)
-        devices = dom_xml.xpath('/domain/devices')[0]
+        self._compat_libvirt_devices(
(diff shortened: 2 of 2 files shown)


mykaul:
@mpolednik - is this the XML for the L1 hosts? For L2 it's all oVirt, right? 
Do you have a concrete example for this use case?

mpolednik:
@mykaul If you mean L1 as first nested layer - yes. This only allows modification of VMs spawned by lago itself.

Very concrete example is testing of hostdev passthrough (= device assignment) - it requires q35 and very specific XML (multiple extra controllers) build to allow vIOMMU. That way, we can assign virtual devices (of L1 host) to oVirt VMs.

I'm not entirely sure of the way it is done right now, as most of the definition from LagoInitFile is simply thrown away though - looking for opinions on how this could be handled better.

lago/prefix.py
```diff
@@ -1182,15 +1182,23 @@ def _copy_deploy_scripts_for_hosts(self, domains):
...
+                new_scripts = self._copy_deploy_scripts(deploy_scripts)
                 self._set_scripts(
                     host_metadata=host_metadata,
                     scripts=new_scripts,
                 )
 
         return domains
 
-    def _copy_delpoy_scripts(self, scripts):
+    def _copy_deploy_scripts(self, scripts):
```
david-caro: there was a tpyo :), good catch.

lago/prefix.py
```diff
@@ -1182,15 +1182,23 @@ def _copy_deploy_scripts_for_hosts(self, domains):
...
-                new_scripts = self._copy_delpoy_scripts(deploy_scripts)
+                if host_metadata.get('custom-xml', None):
+                    host_metadata['custom-xml'] = os.path.expanduser(
+                        os.path.expandvars(
+                            self._copy_deploy_scripts(
+                                [host_metadata['custom-xml']]
+                            )[0]
+                        )
+                    )
+                new_scripts = self._copy_deploy_scripts(deploy_scripts)
```
david-caro: This will throw an exception that might be a bit confusing like 'Script mycustom.xml does not exist', maybe we can do something about that? (just asking if you have a better idea, I'll think about it too)

lago/vm.py
```diff
@@ -223,8 +223,13 @@ def _libvirt_name(self):
...
     def _libvirt_xml(self):
-        with open(_path_to_xml('dom_template.xml')) as xml_fd:
-            dom_raw_xml = xml_fd.read()
+        metadata = self.vm.metadata
+        if metadata and 'custom-xml' in metadata:
+            with open(metadata['custom-xml']) as xml_fd:
+                dom_raw_xml = xml_fd.read()
+        else:
+            with open(_path_to_xml('dom_template.xml')) as xml_fd:
+                dom_raw_xml = xml_fd.read()
```
david-caro: Maybe move this to it's own function, something like self.vm.get_domxml that does the fallback to the default template (or if you want to be fancy, using the python getters and setters for properties, so self.vm.domxml does it magically).

lago/vm.py
```diff
@@ -255,8 +260,16 @@ def _libvirt_xml(self):
...
+        self._compat_libvirt_devices(
+            dom_xml.xpath('/domain/devices')[0])
+
+        return lxml.etree.tostring(dom_xml)
 
+    def _compat_libvirt_devices(self, devices):
+        self._compat_libvirt_disks(devices)
+        self._compat_libvirt_nics(devices)
+
+    def _compat_libvirt_disks(self, devices):
```
david-caro: nice

lago/vm.py
```diff
@@ -318,6 +331,9 @@ def _libvirt_xml(self):
             )
             devices.append(disk)
 
+        return devices
+
+    def _compat_libvirt_nics(self, devices):
```
david-caro: nice

david-caro:
This is something that we wanted to do since we started, to allow using lago on very specific cases even if the functionality to manage the xml was not yet implemented. As @mpolednik points out, maybe we should try to do some merging of the custom xml with the default template or similar when some options are not specified (so it would be like using the defaults from the template xml, and override them with the custom xml), though it might be useful too to completely ignore them and just use the custom one (as in this patch).

Maybe we can go with this and open an issue to implement the other option too somehow.

david-caro:
ci test please

eedri:
ci test please

eedri:
@mpolednik might need a rebase / update of the patch against current master

nvgoldin:
@mpolednik - feel free to re-open if you want to pick it up again.

PR #341, opened 2016-10-10
Collect hosted-engine logs

didib:
Change-Id: Ic3c53c9046078f6c9eeb10b09b24cbdcf608a382
Signed-off-by: Yedidyah Bar David didi@redhat.com

### DIFF
--- a/ovirtlago/virt.py
+++ b/ovirtlago/virt.py
@@ -205,5 +205,7 @@ class HostVM(lago.vm.DefaultVM):
             inherited_artifacts + [
                 '/var/log/vdsm',
                 '/var/log/sanlock.log',
+                '/var/log/ovirt-hosted-engine-setup',
+                '/var/log/ovirt-hosted-engine-ha',
             ]
         )


didib:
Looked at [1] and could not find hosted-engine logs.
I am new to lago. grepped a bit and found this place where it seems I can try adding them. I'd expect it to be in the repo 'ovirt-system-tests', not 'lago'. Also not sure what happens with tests that are not hosted-engine, where these logs will not exist. Will this fail the job? If so, how do I cleanly fix this? Perhaps add a new vm-type? I am just guessing from the code, didn't yet readthedocs.

[1] http://jenkins.ovirt.org/view/oVirt%20system%20tests/job/ovirt_4.0_he-system-tests/lastSuccessfulBuild/artifact/

didib:
Perhaps something like #342

david-caro:
Yep, I think that the other one is the way to go, this is adding the list of artifacts to any host, hosted engine or not.

didib:
Abandoning in favor of #342 (which is merged) and https://gerrit.ovirt.org/65286 .

Issue #47, opened 2016-01-27
docs:Add docs on the configuration files


david-caro:
Moved from https://bugzilla.redhat.com/show_bug.cgi?id=1285729

PR #344, opened 2016-10-13
[docs] lago.conf customization

nvgoldin:
Adding configuration section to the docs. Mainly address changes from https://github.com/lago-project/lago/pull/324.

### DIFF
--- a/docs/Configuration.rst
+++ b/docs/Configuration.rst
@@ -1,0 +1,94 @@
+Configuration
+=============
+The recommend method to override the configuration file is by letting
+lago auto-generate them::
+
+        $ mkdir -p $HOME/.config/lago
+        $ lago generate-config > $HOME/.config/lago/lago.conf
+
+This will dump the current configuration to ``$HOME/.config/lago/lago.conf``,
+and you may edit it to change any parameters. Take into account you should
+probably comment out parameters you don't want to change when editing the file.
+Also, all parameters in the configuration files can be overridden by passing
+command line arguments or with environment variables, as described below.
+
+
+
+lago.conf format
+^^^^^^^^^^^^^^^^
+Lago runs without a configuration file by default, for reference-purposes,
+when lago is installed from the official packages(RPM or DEB),
+a commented-out version of lago.conf(INI format) is installed at
+``/etc/lago/lago.conf``.
+
+In ``lago.conf`` global parameters are found under the ``[lago]`` section.
+All other sections usually map to subcommands(i.e. ``lago init`` command
+would be under ``[init]`` section).
+
+*Example*::
+
+        $ lago generate-config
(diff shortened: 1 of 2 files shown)


nvgoldin:
Issue: https://github.com/lago-project/lago/issues/47

docs/Configuration.rst
```diff
@@ -0,0 +1,92 @@
...
+        > template_repos = /var/lib/lago/repos
+        > # location to store repos
+        > template_store = /var/lib/lago/store
+        > # location to store temp
+
+        $ export LAGO__INIT__TEMPLATE_STORE=/var/tmp
+        $ lago generate-config | grep -A4 "init"
+        > [init]
+        > template_repos = /var/lib/lago/repos
+        > # location to store repos
```
mykaul: It is a bit confusing that the documentation comes AFTER the variable itself.

nvgoldin:
Yes they are under and not before. Had a look at some other configurations and I agree it looks better to have them before, will send a fix(here and more importantly to the code).

nvgoldin:
fixed typos and changed order of print in https://github.com/lago-project/lago/pull/347

docs/Configuration.rst
```diff
@@ -0,0 +1,93 @@
+Configuration
+=============
+The recommend method to override the configuration file is by letting
+lago auto-generate them::
```
nvgoldin: Actually it puts the commented out version in ``/etc/lago/lago.conf`` exactly for that(reference purposes), so maybe I should add that to the documentation. I think its better to output the actual loaded configuration as it might be useful for debugging also(i.e. seeing what is the current configuration for a parameter if you are not sure what is loaded) and not only for editing, in that sense maybe the verb name is a bit misleading(generate-config vs something like show-config). Once someone already edits the config file I assume he will remove what he finds redundant.

As for the documentation, maybe change it to something like:
>The /etc/lago/lago.conf file is commented out and used as an example. You can edit it there and change the global configuration for all users. Also, you may dump the current loaded configuration using ``lago generate-config`` and edit what you need, like this:'

and then the example in the current PR. 

as a side note, another thought I had was to add something like ``lago config`` which will be local to the current used prefix(like ``git config``), this could really simplify stuff for the user(one none-trivial problem is that the prefix exists only after ``lago init``, other than that I think it should be simple)

nvgoldin:
@dcaro - as the actual code implementing it is already in, tell me if you think I should re-factor, as I rather get the docs inside with the next release.

nvgoldin:
Re-factored a little according to the comments. As we are going to release soon, I'll merge it - if you think it still needs iteration i can do it in next version. (rather have some docs out there on the change than nothing..)

nvgoldin:
ci merge please

Issue #354, opened 2016-11-02
IndexError: tuple index out of range when running lago init


dyasny:
Following the documentation at http://lago.readthedocs.io/en/stable/Jenkins_Example.html on a clean RHEL 7.2 with the lago repo and EPEL enabled and lago successfully installed

```
# lago init
Traceback (most recent call last):
  File "/usr/bin/lago", line 10, in <module>
    sys.exit(main())
  File "/usr/lib/python2.7/site-packages/lago/cmd.py", line 777, in main
    parser = create_parser(cli_plugins=cli_plugins, out_plugins=out_plugins, )
  File "/usr/lib/python2.7/site-packages/lago/cmd.py", line 626, in create_parser
    pkgs = e[2]
IndexError: tuple index out of range

```

nvgoldin:
Doesn't look good - can you tell me which version you are using?(lago --version)

dyasny:
```
# lago --version
Traceback (most recent call last):
  File "/usr/bin/lago", line 10, in <module>
    sys.exit(main())
  File "/usr/lib/python2.7/site-packages/lago/cmd.py", line 777, in main
    parser = create_parser(cli_plugins=cli_plugins,
out_plugins=out_plugins, )
  File "/usr/lib/python2.7/site-packages/lago/cmd.py", line 626, in
create_parser
    pkgs = e[2]
IndexError: tuple index out of range
```

dyasny:
# rpm -qa |grep lago
python-lago-0.27.0-1.el7.centos.noarch
lago-0.27.0-1.el7.centos.noarch

nvgoldin:
thanks, this seems related to https://github.com/lago-project/lago/pull/352, might be regression. Can you also check which versions of pbr and stevedore you have installed?
@gbenhaim  - any ideas?

dyasny:
python-stevedore-1.12.0-1.el7ost.noarch
python-pbr-1.8.1-2.el7ost.noarch

If you're inside the RH network I can just let you take a look, ping me on
internal IRC for access details

nvgoldin:
I was able to replicate the issue: it was due to an old version of python-six installed(python-six.noarch 0:1.3.0-4.el7), while stevedore requires: six (>=1.9.0), so a quick workaround was to yum update python-six.
The issue in https://github.com/lago-project/lago/pull/352 is that it masked the error pkg_resources would have printed otherwise:
```
[...]

PR #352, opened 2016-10-27
Allow using lago on systems which runs setevedore 1.1.0

gbenhaim:
1. On some versions of rhel pbr 1.10.0 and stevedore 1.1.0
are the only versions availble. While stevedore 1.1.0
require pbr<1, it can also work with pbr 1.10.0 (tested
on my system). Added an exception handling in order to
bypass the requirements.

2. Changed the way we check for lago's version to suit
the change in [1].

example of the error:

```
traceback (most recent call last):
  File "/usr/bin/lago", line 10, in <module>
    sys.exit(main())
  File "/usr/lib/python2.7/site-packages/lago/cmd.py", line 659, in main
    parser = create_parser(cli_plugins=cli_plugins, out_plugins=out_plugins)
  File "/usr/lib/python2.7/site-packages/lago/cmd.py", line 598, in create_parser
    pkg_info = pkg_resources.require("lago")[0]
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 943, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 834, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (pbr 1.8.2.dev31 (/usr/lib/python2.7/site-packages), Requirement.parse('pbr!=0.7,<1.0,>=0.6'), set(['stevedore']))
```

PR #356, opened 2016-11-03
Fixed exception handling when checking for deps

gbenhaim:
1. Moved deps checking into a function which runs
after the setup of the logger.

2. Minimized the scope of 'except' to catch 'ContextualVersionConflict'
(required only for the 'stevedore' fix:
1e3ebe13fb78f3bdffd95f96b92d6f66b4636fb4).

fix: https://github.com/lago-project/lago/issues/354
part of: https://github.com/lago-project/lago/pull/352

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -621,18 +621,6 @@ def create_parser(cli_plugins, out_plugins):
         '--logdepth', default=3, type=int, help='How many task levels to show'
     )
 
-    try:
-        # Checks that all the deps are installed
-        pkg_resources.require("lago")[0]
-    except pkg_resources.VersionConflict as e:
-        # Hack that allows to run stevedore without checking
-        # for it's dep. it is required for systems running stevedore 1.1.0
-        # and pbr > 1.
-        LOGGER.debug(e.message)
-        pkgs = e[2]
-        if len(pkgs) > 1 or 'stevedore' not in pkgs:
-            raise e
-
     parser.add_argument(
         '--version',
         action='version',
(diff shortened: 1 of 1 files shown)


lago/cmd.py
```diff
@@ -767,6 +756,20 @@ def check_group_membership():
...
+    try:
+        # Checks that all the deps are installed
+        pkg_resources.require("lago")[0]
+    except pkg_resources.ContextualVersionConflict as e:
+        # Hack that allows to run stevedore without checking
+        # for it's dep. it is required for systems running stevedore 1.1.0
+        # and pbr > 1.
+        LOGGER.debug(traceback.format_exc())
+        pkgs = e[2]
+        if len(pkgs) > 1 or 'stevedore' not in pkgs:
```
pilou-: `['stevedore'] != pkgs` seems easier to understand.

lago/cmd.py
```diff
@@ -767,6 +756,20 @@ def check_group_membership():
...
 
+def check_deps():
+    try:
+        # Checks that all the deps are installed
+        pkg_resources.require("lago")[0]
+    except pkg_resources.ContextualVersionConflict as e:
+        # Hack that allows to run stevedore without checking
+        # for it's dep. it is required for systems running stevedore 1.1.0
+        # and pbr > 1.
+        LOGGER.debug(traceback.format_exc())
```
pilou-: `LOGGER.debug(e, exc_info=True)` could be used instead.

Not sure if the message is useful when the exception is re-raised, exception could be logged only when ignored/hidden.

nvgoldin: It shouldn't be in debug level, as his exception should always be logged.
In general - I think we rather use logging.exception('message') for exceptions. see: https://docs.python.org/2/library/logging.html#logging.Logger.exception.

gbenhaim: @pilou- done, thanks for the comment.

gbenhaim:
ci merge please

Issue #345, opened 2016-10-15
[tests] functional tests randomly fail in deploy stage


nvgoldin:
looks like a caching issue on copy/write to VM:

``` bash
11:44:46 ok 36 deploy.basic: setup
11:45:22 not ok 37 deploy.basic: deploy
11:45:22 # (from function `helpers.diff' in file tests/functional/helpers.bash, line 129,
11:45:22 #  from function `helpers.diff_output' in file tests/functional/helpers.bash, line 116,
11:45:22 #  in test file tests/functional/deploy.bats, line 50)
11:45:22 #   `helpers.diff_output "$FIXTURES/nicefile"' failed
11:45:22 # ~/tests/functional/fixtures/deploy ~
11:45:22 # RUNNING:lago deploy
11:45:22 # --output--
11:45:22 # current session does not belong to lago group.
11:45:22 # @ Deploy environment: 
11:45:22 #   # [Thread-1] Deploy VM lago_functional_tests_vm01: 
11:45:22 #     * [Thread-1] Wait for ssh connectivity: 
11:45:22 #     * [Thread-1] Wait for ssh connectivity: Success (in 0:00:22)
11:45:22 #     * [Thread-1] Run script _home_jenkins_workspace_lago_master_check-merged-fc24-x86_64_lago_tests_functional_fixtures_deploy_.lago_default_.._.._scripts_create_nicefile.sh: 
11:45:22 #     * [Thread-1] Run script _home_jenkins_workspace_lago_master_check-merged-fc24-x86_64_lago_tests_functional_fixtures_deploy_.lago_default_.._.._scripts_create_nicefile.sh: Success (in 0:00:01)
11:45:22 #     * [Thread-1] Run script _home_jenkins_workspace_lago_master_check-merged-fc24-x86_64_lago_tests_functional_fixtures_deploy_.lago_.._scripts_create_uglyfile.sh: 
11:45:22 #     * [Thread-1] Run script _home_jenkins_workspace_lago_master_check-merged-fc24-x86_64_lago_tests_functional_fixtures_deploy_.lago_.._scripts_create_uglyfile.sh: Success (in 0:00:01)
11:45:22 #   # [Thread-1] Deploy VM lago_functional_tests_vm01: Success (in 0:00:25)
11:45:22 # @ Deploy environment: Success (in 0:00:25)
11:45:22 # ---
11:45:22 # "0" == "0"
11:45:22 # Checking that the nicefile was properly created
[...]

PR #365, opened 2016-11-19
For ease of use, connect to a host even if only the suffix is given.

mykaul:
If a host name for a shell command is not found, try to look for it
based on its suffix. This allows 'lagocli shell engine' to work,
for example, instead of 'lagocli shell lago-basic-suite-master-engine'

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -339,12 +339,20 @@ def do_shell(prefix, host, args=None, **kwargs):
     try:
         host = prefix.virt_env.get_vm(host)
     except KeyError:
-        LOGGER.error('Unable to find VM %s', host)
-        LOGGER.info(
-            'Available VMs:\n\t' +
-            '\n\t'.join(prefix.virt_env.get_vms().keys())
-        )
-        raise
+        ssh_host = None
+        for possible_host in prefix.virt_env.get_vms():
+            if possible_host.endswith(host):
+                ssh_host = prefix.virt_env.get_vm(possible_host)
+                break
+        if ssh_host:
+            host = ssh_host
+        else:
+            LOGGER.error('Unable to find VM %s', host)
+            LOGGER.info(
+                'Available VMs:\n\t' +
+                '\n\t'.join(prefix.virt_env.get_vms().keys())
+            )
+            raise
 
     if not host.alive():
         raise RuntimeError(


ifireball:
This is IMO very strange and case-specific behaviour (Prefix-based search would better meet user expectations IMO, but even that is risky given possible collisions and typos). I understand this comes from the desire to do 'lago shell engine' in an OST environment. Maybe it would be better to just change the host names in OST to not be so long? I don't think the decision to pretend the suit name to the host names there was a good or a needed decision.

mykaul:
It's not a very case-specific behavior, as I assume prefix is one of the main ideas in Lago.
Even if it is a very case specific behavior, it is for the most widely used use-case of Lago.

I do hope to run multiple prefixes concurrently, so there is an advantage of naming the VMs fully.

And the worst that can happen is that you SSH to the wrong VM. No big deal. I still think it's worthwhile addition.

ifireball:
VMs in different prefixes run in different networks - name collisions between prefixes are impossible, so there shouldn't be a need to give VMs long names from within the prefix.
Prefix-level network isolation may be the single most important feature of Lago, it that isn't working we need to solve that, not come up with ways to make workarounds convenient.

mykaul:
Any ideas why CI fails while 'make check' locally passes?

ifireball:
Looks like a pyflakes failure:
```
07:26:33 ./lago/templates.py:264:1: E305 expected 2 blank lines after class or function definition, found 1
```
You didn't change that file so maybe someone else broke master.

If you want to really do what CI does - Run check_patch.sh with mock_runner.

nvgoldin:
:+1: on the patch, I think we can improve it later if needed. Lets say this is for sure an enhancement than the current situation. 
about the CI failure - looks like something similar to https://gitlab.com/pycqa/flake8/issues/246, quick googling discovered many new failures introduced this month by new version of pycodestyle pushed, which flake8 depends on. I suggest trying to fix the minor style issue, and if it won't work we'll need to pin the versions(hopefully they are still available) in the spec file/mock configurations.

mykaul:
ci test please

mykaul:
@nvgoldin yapf on my laptop and on CI disagree. Works OK on my laptop (changed according to latest yapf recommendations) and CI wants a different settings?

nvgoldin:
@ykaul, for this patch? the fail is not on yapf(at least the last run):
>19:28:22 scripts/check_style.sh
19:28:22 yapf 0.14.0
19:28:22 running yapf on the following files:
19:28:22 lago/cmd.py
19:28:24 yapf style check: OK

The failure in http://jenkins.ovirt.org/job/lago_master_check-patch-fc23-x86_64/365/  looks(again) like: https://github.com/lago-project/lago/issues/345.

nvgoldin:
ci test please

mykaul:
ci merge please

nvgoldin:
after using it for some time(under none-OST VMs) - I agree its really none-standard, I think we should move to 'grep' like behaviour.

mykaul:
@nvgoldin - I actually use it on a daily basis and enjoy it very much. If you have a better approach, sure - please suggest it / send a patch. If you really suffer from it, it can of course be reverted. I'll survive without it.

nvgoldin:
@mykaul, of course its better than what we had before, just noticed now that the behaviour is somewhat strange. I would go for:
1. Match substring, maybe limit it to >2 chars.
2. if matches: print what was matched to stdout, and use it. else fail and print  possible VMs.

Will try to send a patch later.

ifireball:
What happens on multiple matches? I suppose we need to print them all and not connect?

Also - we should probably not print the match to stdout - because it would break scripts that might depend on parsing "lago shell" output.

I see to possible solutions:

1. Print to stdout only when matching substring - not when fully matching string was given (But also give an option to suppress output even then)
2. Print to stderr instead of stdout (might not be appropriate...)

mykaul:
On Tue, Dec 6, 2016 at 12:12 PM, Barak Korren <notifications@github.com>
wrote:

> What happens on multiple matches? I suppose we need to print them all and
> not connect?
>

Are we being theoretical here, or is it a real use case?
Yes, if you have two hosts:
first_host
second_host

and you try to connect to host, it'll always go to 'first_host'. So either
use a full name, or 'd_host' if you want to ensure it's the 2nd host. This
is why it's a convenience feature that you need to properly use.


> Also - we should probably not print the match to stdout - because it would
> break scripts that might depend on parsing "lago shell" output.
>

Again, a theoretical item - but my patch did not change the behavior anyway.


> I see to possible solutions:
>

I don't see the problems yet.

PR #366, opened 2016-11-21
Switch to optional freeze the FS before logs extraction.

mykaul:
More importantly, by default, the call will not freeze the FS.
There is no reason to do it, for two reasons:
1. We open the FS in read-only - so no corruption can happen.
2. The system is quite stable - we already passed whatever
test we passed or failed - all logs are in the system already
(especially as in ovirt-system-tests we complete the suite after
failure)

This may prevent issues when the hosts are disconnected due to a
long freeze

### DIFF
--- a/lago/vm.py
+++ b/lago/vm.py
@@ -408,14 +408,16 @@ class LocalLibvirtVMProvider(vm.VMProviderPlugin):
             self._reclaim_disks()
             self.vm._spec['snapshots'][name] = snap_info
 
-    def _extract_paths_live(self, paths):
-        self.vm.guest_agent().start()
-        dom = self.libvirt_con.lookupByName(self._libvirt_name())
-        dom.fsFreeze()
+    def _extract_paths_live(self, paths, freeze=False):
+        if freeze:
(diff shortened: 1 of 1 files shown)


eedri:
Failing on:
19:23:32 ##!
19:23:32 ./lago/prefix.py:745:27: F999 dictionary key 'name' repeated with different values
19:23:32 ./lago/prefix.py:746:27: F999 dictionary key 'name' repeated with different values
19:23:32 ./lago/templates.py:264:1: E305 expected 2 blank lines after class or function definition, found 1
19:23:32 make: *** [check-local] Error 1
19:23:32 ~~~~~~~~~~~~~~~~~~~~

mykaul:
ci test please

mykaul:
@nvgoldin @gbenhaim @david-caro : sent an email to lago-devel - yapf and flake8 disagree with each other, and I'm stuck on this (which has NOTHING to do with this patch!)

gbenhaim:
+1

gbenhaim:
ci merge please

Issue #319, opened 2016-09-15
Remote repository synchronization should be more reliable


sandrobonazzola:
Remote repository synchronization should be more reliable.
In the following logs from http://jenkins.ovirt.org/job/vdsm_4.0_check-merged-fc23-x86_64/146/console 
- It says that a first reposync run completed with success.
- It then advise that sometime reposync fails and try again but previous sync succeeded.
- It resync and finsih with success.
- It the says it failed and it tries again.
- It says it finished with success and then fail again.

the repository synchronization should be more reiliable and use all available mirrors.

`0:06:53.074   # Syncing remote repos locally (this might take some time): 
00:06:53.074     \* Acquiring lock for /var/lib/lago/reposync/repolock: 
00:06:53.075     \* Acquiring lock for /var/lib/lago/reposync/repolock: Success (in 0:00:00)
00:06:53.077     \* Running reposync: 
00:06:58.904     \* Running reposync: Success (in 0:00:05)
00:06:58.905     \* Due to bug https://bugzilla.redhat.com/show_bug.cgi?id=1332441 sometimes reposync fails to update some packages that have older versions already downloaded, will remove those if any and retry
00:06:59.019     \* Rerunning reposync: 
00:06:59.839     \* Rerunning reposync: Success (in 0:00:00)
00:06:59.840     \* Failed to run reposync again, that usually means that some of the local rpms might be corrupted or the metadata invalid, cleaning caches and retrying a second time
00:06:59.896     \* Rerunning reposync a last time: 
00:07:00.837     \* Rerunning reposync a last time: Success (in 0:00:00)
00:07:00.838     \* reposync command failed with following output: 
local-vdsm-build-fc23                                       | 2.9 kB  00:00  
00:07:00.839 
local-vdsm-build-fc23/primary_db                            |  17 kB  00:00  
00:07:00.839 
ovirt-4.0-snapshot-fc23                                     | 2.9 kB  00:00  
00:07:00.839 
ovirt-4.0-snapshot-fc23/primary_db                          |  95 kB  00:00  
00:07:00.840 
ovirt-4.0-snapshot-static-fc23                              | 2.9 kB  00:00  
00:07:00.840 
ovirt-4.0-snapshot-static-fc23/primary_db                   | 8.0 kB  00:00  
00:07:00.840 
ovirt-release40-4.0.0-1.noarch FAILED  
00:07:00.840 
[...]

PR #369, opened 2016-11-29
make reposync more verbose and fix bz 1399235

nvgoldin:
Until now we tried deducing the RPM name in reposync errors
from the line printing 'no more mirrors to try', however in the BZ
1399235, it is actually the rpm spec file 'Release' parameter that
is printed.
This could have been due to a change in yum/yum-utils, or we are
now seeing a different error.

This patch attempts to make the 'reposync' stage more robust and
verbose:

1. New regex lookups were added according to the BZ, the old regex
(probably due to yum/yum-utils changes) is looking for a different
error and also sometimes matched incorrect RPM names(which starting with
'^M' sequence).
It attempts to look RPM names according to the fedora guidelines:
https://fedoraproject.org/wiki/How_to_create_an_RPM_package

2. Defected packages lookup will be done inside python instead
of GNU find and when deleting a file, it will be logged and
also printed to stdout.

3. reposync will be called for each repoid separately, this will
allow to easily trace which repository caused the error.

bz: https://bugzilla.redhat.com//show_bug.cgi?id=1399235

@sandrobonazzola  - should probably close https://github.com/lago-project/lago/issues/319, or at least improve it.

### DIFF
--- a/ovirtlago/reposetup.py
+++ b/ovirtlago/reposetup.py
@@ -20,6 +20,7 @@
 
 import functools
 import re
+import itertools
 import shutil
 import logging
 import os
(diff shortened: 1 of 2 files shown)


ovirtlago/reposetup.py
```diff
@@ -72,33 +73,55 @@ def wrapper(*args, **kwargs):
...
-        https://bugzilla.redhat.com/show_bug.cgi?id=1332441
+        https://bugzilla.redhat.com//show_bug.cgi?id=1399235
+        https://bugzilla.redhat.com//show_bug.cgi?id=1332441
+
     """
-    LOGGER.warn(
-        'Due to bug https://bugzilla.redhat.com/show_bug.cgi?id=1332441 '
-        'sometimes reposync fails to update some packages that have older '
-        'versions already downloaded, will remove those if any and retry'
+    if any([len(repo_path) == 0, len(reposync_out) == 0]):
```
mykaul: what's wrong with 'or' ?

ovirtlago/reposetup.py
```diff
@@ -72,33 +73,55 @@ def wrapper(*args, **kwargs):
...
+        LOGGER.warning(
```
mykaul: LOGGER.warning or LOGGER.warn ?

ovirtlago/reposetup.py
```diff
@@ -72,33 +73,55 @@ def wrapper(*args, **kwargs):
...
+    )
+    wrong_release = re.compile(r'(?P<package_name>' + rpm_regex + ') FAILED')
+    packages = set(
+        itertools.chain(
+            wrong_version.findall(reposync_out),
+            wrong_release.findall(reposync_out)
+        )
+    )
+    count = 0
+    LOGGER.debug(
```
mykaul: Worth separating version and release? perhaps it'll help debugging the issue later?

ovirtlago/reposetup.py
nvgoldin: fixed.

ovirtlago/reposetup.py
nvgoldin: actually I did that by accident, but it turns out ``logging.warn`` is going to be deprecated, so I'll keep it that way. http://bugs.python.org/issue13235

ovirtlago/reposetup.py
nvgoldin: As per the logging - I don't think it matters, because eventually the user will see packages were removed in the output(in stdout, not only in the the log file), and this is a reposync bug, so there is nothing he can really debug here.
About the regex split few lines above(one for each bug) - I think that will help us debug/improve the code itself it in the future, if we encounter new issues with reposync.

nvgoldin:
ci test please

nvgoldin:
ci test please

nvgoldin:
ci merge please

ovirtlago/reposetup.py
```diff
@@ -107,42 +130,48 @@ def sync_rpm_repository(repo_path, yum_config, repos):
...
+        '--cachedir=%s/cache' % repo_path
+    ]
+    with LogTask('Running reposync'):
+        for repo in repos:
+            with LockFile(lock_path, timeout=180):
+                reposync_cmd = reposync_base_cmd + ['--repoid=%s' % repo]
+                ret, out, _ = run_command(reposync_cmd)
+                if not ret:
+                    LOGGER.debug('reposync on repo %s: success.' % repo)
+                    return
```
mykaul: This should be 'continue'

ovirtlago/reposetup.py
```diff
@@ -107,42 +130,48 @@ def sync_rpm_repository(repo_path, yum_config, repos):
...
+
+                LOGGER.info('repo: %s: failed, re-running.', repo)
+                _fix_reposync_issues(
+                    reposync_out=out, repo_path=os.path.join(repo_path, repo)
+                )
+                ret, _, _ = run_command(reposync_cmd)
+                if not ret:
+                    return
```
mykaul: This should be continue

mykaul:
Note that calling reposync to each repository separately probably will slow down considerably the sync process. Need to think if this is what we really want - at least let's do them in parallel threads?

nvgoldin:
My bad, thanks for https://github.com/lago-project/lago/pull/374.

>Note that calling reposync to each repository separately probably will slow down considerably the sync process. Need to think if this is what we really want - at least let's do them in parallel threads?

I had a look at reposync code and indeed it first [collects](https://github.com/rpm-software-management/yum-utils/blob/master/reposync.py#L206) a list of all packages from all enabled repositories and then [downloads all of them](https://github.com/rpm-software-management/yum-utils/blob/master/reposync.py#L306), so yes - we will loose speed here. (originally I thought it was less efficient than that..)

Considering reposync's failure rate, I still think its much better to have it run separately on each repository for logging and debugging, especially when RPMs are removed, so you get a message in the log exactly which repository is misbehaving. I'll try making it to parallel, if it becomes too complicated and speed turns out to be significant, I'll revert it back.

PR #371, opened 2016-12-05
Fix SSH with public key and with username

mykaul:
1. We were never using public key auth.!
We've always supplied the user's password,
while the key password is actually ''" (empty string).

2. Added not to look for keys in the system and not connect to the agent.
Not needed.

3. Switch SSH with username to use any username passed to it. The default
is still root of course. Now it can also accept a password.

4. Move the call to initialize the paramiko object outside the loop.

### DIFF
--- a/ovirtlago/testlib.py
+++ b/ovirtlago/testlib.py
@@ -22,7 +22,7 @@ import functools
 import logging
 import os
 import time
-
+import unittest.case
 import nose.plugins
 from nose.plugins.skip import SkipTest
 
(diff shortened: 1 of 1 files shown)


mykaul:
The whole public key infrastructure is completely broken - mainly because it'll never get used. There is no chance the VM will not have a root or SSH password. And they are actually taking one from another.

See code mainly in lago/plugins/vm.py (specifically, _normalize_spec() and root_password() functions are of interest):
1. There is also ssh-password, as it is set in lago/cmd.py by default. 
2. There is never a root_password in the spec (unless explicitly set.). It is (probably always) in the metadata of the image - but no one bothers to read it from it?
3. There is a default_root_password set in lago/cmd.py - don't think it's used anywhere in the code.
4. If you look at root_password() - it is trying to read it from the spec (which is never there) and if it fails, it copies it from spec's ssh-password. Of course, earlier ssh-password was set in _normalize_spec()

So all in all:
1. A bit of a mess.
2. No way public key would work, as we always have a password.

@nvgoldin  , @david-caro  - thoughts?

I'll probably do the minimal changes (2-4 above) and forget about the rest for the time being, unless we can properly fix this.

PR #376, opened 2016-12-08
Reduce boot time

mykaul:
1. Remove the grub boot menu in virt-sysprep phase
2. Remove the QEMU boot menu in domain template XML
3. Wait 5 seconds till the VM is potentially ready to
receive SSH connections

### DIFF
--- a/lago/sysprep.py
+++ b/lago/sysprep.py
@@ -117,6 +117,14 @@ def config_net_interface_dhcp(iface, hwaddr):
     )
 
 
+def edit(filename, expression):
+    editstr = '%s:""%s""' % (filename, expression)
+    return (
+        '--edit',
+        editstr,
+    )
+
+
 def sysprep(disk, mods, backend='direct'):
     cmd = ['virt-sysprep', '-a', disk, '--selinux-relabel']
     env = dict(os.environ.copy(), LIBGUESTFS_BACKEND=backend)
--- a/lago/vm.py
+++ b/lago/vm.py
@@ -94,7 +94,11 @@ class LocalLibvirtVMProvider(vm.VMProviderPlugin):
         super(LocalLibvirtVMProvider, self).start()
         if not self.defined():
             with LogTask('Starting VM %s' % self.vm.name()):
-                self.libvirt_con.createXML(self._libvirt_xml())
+                dom = self.libvirt_con.createXML(self._libvirt_xml())
+                if not dom:
+                    raise RuntimeError(
+                        'Failed to create Domain: %s' % self._libvirt_xml()
+                    )
 
     def stop(self):
         super(LocalLibvirtVMProvider, self).stop()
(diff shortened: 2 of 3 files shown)


lago/prefix.py
```diff
@@ -1225,6 +1226,7 @@ def _deploy_host(self, host):
                 return
 
             with LogTask('Wait for ssh connectivity'):
+                time.sleep(5)
```
david-caro: This will make you wait 5 seconds every time you want to deploy the host no? Maybe we can do a first try before the sleep? It seems to me also that this belongs more into the `wait_for_ssh` function, or maybe to the start one if it's only needed when booting up the vm.

lago/sysprep.py
```diff
@@ -117,6 +117,14 @@ def config_net_interface_dhcp(iface, hwaddr):
     )
 
 
+def edit(filename, expression):
```
david-caro: Nice

lago/sysprep.py
```diff
@@ -117,6 +117,14 @@ def config_net_interface_dhcp(iface, hwaddr):
...
+    editstr = '%s:""%s""' % (filename, expression)
```
david-caro: Just to make sure, here there are two pairs of double quotes, is that what you intended?

lago/prefix.py
mykaul: In practice, no VM loads in less than 5 seconds, so why bother it with useless SSH requests?
You should not put it inside the wait_for_ssh(), as this wait can be used elsewhere. It's only on the very first call, when the VM loads.

lago/sysprep.py
mykaul: Yes, it is. virt-edit specifically mentions that it should be used to ensure escaping.

lago/prefix.py
david-caro: Well, the issue is that the deploy command is not used only on boot, but also whenever you want no? It's a command of it's own.

Sorry I might not have explained well, let me rephrase, we should do one of:
   * Instead of doing the sleep first, do a first try, and then sleep (if keeping it in the deploy step).
   * Move the sleep to wherever we do the boot (probably the start step).

mykaul: OK, so the caller of env_start essentially will do the sleep, before calling env_deploy?

mykaul: @david-caro - see latest revision. I've removed the wait - moved it to ovirt-system-tests - after it starts the VM, ./run_suite.sh simply sleep for 5 seconds. Brings us the same effect, without touching the deploy part in Lago.

mykaul:
ci test please

lago/vm.py
```diff
@@ -94,7 +94,11 @@ def start(self):
         super(LocalLibvirtVMProvider, self).start()
         if not self.defined():
             with LogTask('Starting VM %s' % self.vm.name()):
-                self.libvirt_con.createXML(self._libvirt_xml())
+                dom = self.libvirt_con.createXML(self._libvirt_xml())
+                if not dom:
+                    raise RuntimeError(
+                        'Failed to create Domain: %s' % self._libvirt_xml()
+                    )
```
david-caro: This is how you ensure that its created! :) 👍

david-caro:
Looks good to me!

lago/vm.py
```diff
@@ -123,6 +127,10 @@ def bootstrap(self):
                         ),
                         sysprep.set_iscsi_initiator_name(self.vm.iscsi_name()),
                         sysprep.set_selinux_mode('enforcing'),
+                        sysprep.edit(
+                            "/boot/grub2/grub.cfg",
+                            "s/set timeout=5/set timeout=0/g"
+                        ) if self.vm.distro() == 'el7' else '',
```
nvgoldin: is there a reason to restrict it to el7 only? (at least on my laptop with fc24 it looks like the same substitution would work)

mykaul:
ci test please

mykaul:
ci merge please

PR #377, opened 2016-12-10
ovirtlago: log skipped tests to stdout

nvgoldin:
Tests which are stopped by raising SkipTest exception, will be logged to stdout. Until now they were only logged in the test report XML.

Example output from OST:
```bash
  # list_glance_images: Success (in 0:00:03)
  # add_secondary_storage_domains: 
  # add_secondary_storage_domains: Success (in 0:02:22)
  # import_templates: 
    * SKIPPED: Exported domain generation not supported yet
  # import_templates: Success (in 0:00:00)
  # add_quota_storage_limits: 
  # add_quota_storage_limits: Success (in 0:00:00)
....
  # vm_migrate: Success (in 0:00:05)
  # snapshot_live_merge: 
    * SKIPPED: 
  # snapshot_live_merge: Success (in 0:00:00)

```

side-note: another problem I noticed while handling this is that if a none-SkipTest exception was thrown, subsequent SkipTest exceptions will loose their description, in stdout they will appear only with ``SKIPPED: ``. I think the reason is mishandling of the exceptions in  [test_sequnce_gen](https://github.com/lago-project/lago/blob/master/ovirtlago/testlib.py#L130-#L148) function. I'll address that in a different PR.

### DIFF
--- a/ovirtlago/testlib.py
+++ b/ovirtlago/testlib.py
@@ -22,7 +22,7 @@ import functools
 import logging
 import os
 import time
-
+import unittest.case
 import nose.plugins
 from nose.plugins.skip import SkipTest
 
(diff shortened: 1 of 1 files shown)


mykaul:
I did not review the code thoroughly (just one item there). Main conclusion is that we need to move to PyTest ;-)

but the whole patch is important and I'd be happy to see it in.

nvgoldin:
> Main conclusion is that we need to move to PyTest ;-)

👍 I agree. The nose bug I linked in the comment probably not going to be fixed.

Also when we re-factor I think it would be better to wrap as little as possible the 'pytest' command and let it do its trick without us maintaining configurations in the code(let it be an external config file).

nvgoldin:
ci test please

nvgoldin:
ci test please

nvgoldin:
ci merge please

nvgoldin:
ci merge please

PR #379, opened 2016-12-11
Update firewall configuration instructions

rgolangh:
The local repo lago creates needs to be accessible from intenal connection. 
Machines that uses iptables and not firewalld must have that set.

### DIFF
--- a/docs/Installation.rst
+++ b/docs/Installation.rst
@@ -86,4 +86,7 @@ refreshes it's permissions and picks up any newly created users::
 
     $ sudo service libvirtd restart
 
+Iptables must allow internal connection to the internal repository lago publishes:
 
+    $ echo "-A INPUT -p tcp --dport 8585 -j ACCEPT" >> /etc/sysconfig/iptables
+    $ service iptables reload


rgolangh:
should I extend lago.spec.in instead? I see it already updates firewalld in case it is installed but doesn't handl iptables

nvgoldin:
@rgolangh   👍  on patching the spec file. even though:
1. we are not supposed to setup ip rules in the sepc file according to packaging standards. 
2. we plan to move this requirment to OST packaging(be it pip or ost).

BUT - as in the meanwhile we do it and its here , it would be better to have a working spec file for none-firewalld installations. 

btw, this happened without firewalld enabled or after upgrade? (https://github.com/lago-project/lago/issues/380)

About this patch -  you mind sending it again to OST?
I think it would better suite there, as this step is only required for 'reposync' stage in ovirtlago(lago is fully functional without this rule)

http://ovirt-system-tests.readthedocs.io/en/latest/docs/general/installation.html, 
which is generated from:
https://gerrit.ovirt.org/gitweb?p=ovirt-system-tests.git;a=blob;f=docs/general/installation.md;h=b8177360c1d33346821e123514ec6a8b6f34b5bc;hb=HEAD

nvgoldin:
closing as this(important) issue should be sent to OST docs.

PR #383, opened 2016-12-13
use LIBGUESTFS_BACKEND variable if set

nvgoldin:
Until now we always used 'direct' in virt-sysprep command.
With this change:
1. 'direct' will be used if LIBGUESTFS_BACKEND is not set.
2. 'LIBGUESTFS_BACKEND' value will be used if set.

According to the [docs](http://libguestfs.org/guestfs.3.html#backend) libguestfs is usally compiled with the 'direct' backend, but not always. So setting this to 'direct' unless the user explicitly asked something else should be safe.


nvgoldin:

nvgoldin:
ci merge please

nvgoldin:
ci merge please

Issue #386, opened 2016-12-15
Add ability to execute commands on the guest directly with virsh console


nvgoldin:
This can be useful when sshd is not available. The none-trivial part would be using pexpect or similar to pass in the credentials.

ifireball:
I think we already have a 'lago console' command.

mykaul:
I was actually hoping to get rid of the video device in favor of a serial console of some sort ;-)

nvgoldin:
@ifireball  - yep but you still need to pass the root/password, and I don't know of any automated way to do it with virsh, so it will require some parsing.

ifireball:
@nvgoldin chatting with 'login' over the serial console sounds like a fragile approach to me. I'd rather we have other channels, like for e.g. the guest agent, before falling back to this.

nvgoldin:
@mykaul, @ifireball  - I agree, what other options we have? if the guest-agent is running then sshd is probably also running, so we can just use sshd. This actually came to my attention as a fall-back to start the guest-agent itself(or at least verify its running) when we don't have sshd :)

ifireball:
@nvgoldin there are plenty of reasons why there could be a guest agent running with sshd inaccessible. For example, the sshd port could be blocked, or the image could have bad IP configuration.

PR #388, opened 2016-12-17
Refactor collect command to use SCP

nvgoldin:
1. Change the default method for extracting files with 'lago collect'
command to SCP. Extracting using libguestfs becomes the fallback
when SSH is not available.
2. Introduce a new option '--no-skip' which will force not skipping
missing paths. If used it will raise a ``ExtractPathNoPathError``
exception if any of the paths does not exist on the host.
3. Drop the option of using fsFreeze/FsThaw when extracting
files with libguestfs.
4. Redirect 'lago ovirt collect' to 'lago collect' and issue a warning
if used directly.
5. Change the functional tests accordingly.

Some more notes:
1. As SCP becomes the first priority - there is no point in keeping the fsFreeze/fsThaw method, because it requires the guest-agent to be running, but we are only able to ensure it is running with SSH(see https://github.com/lago-project/lago/issues/386). 
2. I suggest dropping ``lago ovirt collect`` entirely here in favour of ``lago collect``, as far as I can tell, it is just code duplication. The logs to collect(aka 'artifacts') are defined in the VM. For now I only added a warning and redirection to ``lago collect``.
3. Making the default behaviour ignoring missing paths, is not backward-incompatible IMO(as discussed in https://github.com/lago-project/lago/pull/211) because until now, when extracting paths with libguestfs it did not fail on missing paths as it was supposed to.
4. The SCP module masks 'IOError' with its own ``SCPException``, so I had to parse the exception message, this works but is a little fragile.

### DIFF
--- a/lago/plugins/vm.py
+++ b/lago/plugins/vm.py
@@ -34,7 +34,7 @@ import os
 import warnings
 from abc import (ABCMeta, abstractmethod)
 
-from scp import SCPClient
+from scp import SCPClient, SCPException
 
 from .. import (
     utils,
@@ -48,11 +48,15 @@ LOGGER = logging.getLogger(__name__)
 LogTask = functools.partial(log_utils.LogTask, logger=LOGGER)
 
 
-class VMErrror(Exception):
+class VMError(Exception):
     pass
 
 
-class ExtractPathError(VMErrror):
+class ExtractPathError(VMError):
+    pass
+
+
+class ExtractPathNoPathError(VMError):
     pass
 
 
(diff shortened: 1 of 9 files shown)


nvgoldin:
ci test please

lago/plugins/vm.py
```diff
@@ -179,35 +183,52 @@ def interactive_console(self):
...
+            :exc:`~lago.plugins.vm.ExtractPathError`: on all other failures.
         """
-        if self.vm.alive() and self.vm.ssh_reachable():
-            self._extract_paths_scp(paths=paths)
-        elif self.vm.alive():
-            raise ExtractPathError(
-                'Unable to extract logs from alive but unreachable host %s. '
-                'Try stopping it first' % self.vm.name()
-            )
+        if self.vm.alive() and self.vm.ssh_reachable(
```
mykaul: If it's alive but not reachable via SSH, don't we won't to fallback to the old method of fetching files?

nvgoldin: Yep - its here: https://github.com/lago-project/lago/pull/388/files/57b0f421773ceb2b144799bcc1a34949d8ba2d96#diff-e1c0bb7009432f911230e11222d155edR203

The ``plugins/vm.py`` is supposed to be the default implementation(which only uses SCP).
in ``lago/vm.py`` it uses this method and then falls back to libguestfs on failures.

nvgoldin:

nvgoldin:
ci merge please

mykaul:
ci merge please

nvgoldin:
ci test please

nvgoldin:
ci test please

nvgoldin:
ci merge please

Issue #389, opened 2016-12-18
functional tests became extremely slow on el7 and fail CI


nvgoldin:
happens locally and:

http://jenkins.ovirt.org/job/lago_master_check-patch-el7-x86_64/417/
http://jenkins.ovirt.org/job/lago_master_check-patch-el7-x86_64/416/

nvgoldin:
I am guessing it is related to CentOS 7.3 being pushed and misbehaving new packages under mock.

nvgoldin:
So the root cause is that under mock(might be in 7.3 in general) virt-sysprep is stuck on:
```
4 0
SeaBIOS (version 1.9.1-5.el7)
Booting from ROM...
Probing EDD (edd=off to disable)... 
```
adding ``export LIBGUESTFS_APPEND=edd=off`` resolves the issue, still testing if this is mock specific.

nvgoldin:
https://bugzilla.redhat.com/show_bug.cgi?id=1404287

ifireball:
Looking at the logs you've places in the bug report - it seems you are not using qemu-kvm-ev. Why is that?

nvgoldin:
@ifireball  - Good point to check what we install under mock. In the bugzilla - I wanted to check the bug is reproducible on "plain" CentOS 7.3 VM taken directly from virt-builder after executing ``yum install libguestfs-tools``.

nvgoldin:
fixed by https://github.com/lago-project/lago/issues/391

PR #391, opened 2016-12-18
automation: configure KVM and libguestfs parameters

nvgoldin:
1. Fix for #389
2. Configure libguestfs tmp dir to be /dev/shm for potential
speed improvements in the functional tests.
3. Ensure /dev/kvm device exists under mock(used by libguestfs)
4. Remove export LIBGUESTFS_DEBUG=1 LIBGUESTFS_TRACE=1 from functional
tests

### DIFF
--- a/automation/common.sh
+++ b/automation/common.sh
@@ -3,6 +3,21 @@
 # Common functions for the scripts
 #
 
+
+set_guestfs_params() {
+    # see: https://bugzilla.redhat.com/show_bug.cgi?id=1404287
+    export LIBGUESTFS_APPEND="edd=off"
+    # make libguestfs use /dev/shm as tmpdir
+    export LIBGUESTFS_CACHEDIR="/dev/shm"
+    export LIBGUESTFS_TMPDIR="/dev/shm"
+
+    # ensure KVM is enabled under mock
+    ! [[ -c "/dev/kvm" ]] && mknod /dev/kvm c 10 232
+
+    # un-comment this to debug LIBGUESTFS
+    # export LIBGUESTFS_DEBUG=1 LIBGUESTFS_TRACE=1
+}
+
 code_changed() {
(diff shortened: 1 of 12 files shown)


automation/common.sh
```diff
@@ -3,6 +3,16 @@
 # Common functions for the scripts
 #
 
+
+set_guestfs_params() {
+    # see: https://bugzilla.redhat.com/show_bug.cgi?id=1404287
+    export LIBGUESTFS_APPEND="edd=off"
+    # make libguestfs use /dev/shm as tmpdir
+    export LIBGUESTFS_CACHEDIR="/dev/shm"
```
mykaul: 1. I was surprised it did not add as much as I hoped when I tried it manually previously - in Lago itself (in the sysprep stage). But why not change it there as well (the same way we control the backend).
2. I have a suspicion other parameters should also go there:
LIBGUESTFS_TMPDIR
XDG_RUNTIME_DIR
TMPDIR

nvgoldin: 1. Actually looking at it now it uses /tmp as default which is tmpfs usually, so I wouldn't expect to see great improvements here. I don't think we should enforce any of LIBGUESTFS_* - if the user sets 'LIBGUESTFS_*' it should always be used(see https://github.com/lago-project/lago/pull/383). Using /dev/shm is not that standard.
2. Trying again with LIBGUESTFS_TMPDIR.. lets see if it yields better results. Next thing is to disable LIBGUESTFS_DEBUG in the functional tests.

nvgoldin: No significant improvements :/ trimmed around 30 seconds from the functional tests here(of check-path) to ~490 seconds.
I suggest we keep it like this, as it probably can't harm.

nvgoldin:

nvgoldin:
ci merge please

PR #398, opened 2016-12-19
Choose the same CPU models

mz-pdm:
CPU model for Lago VMs is passed through from the host and the cluster CPU
family is set based on that CPU model.  This approach has some problems:

- CPU model is set according to the host CPU, which means that the tests are
  not always run in the same environment.

- The environment deployment may fail due to invalid CPU family of the cluster.
  Although the family corresponds to the host CPU, kernel may remove some CPU
  flags in nested virtual machines (when the Lago host is actually a VM).  Then
  some CPU flags required for the given CPU family (such as Haswell-NoTSX) may
  be missing in the Lago virtual machines.

- Newer CPU families are not recognized until the table of CPU families is
  updated.

This patch fixes the problems by selecting a minimum reasonable CPU model for
the given hardware platform.  Westmere is selected unless older or non-Intel
hardware is used.  This way the same CPU is always selected on modern Intel
hardware, which makes the testing environment more deterministic and the
problems with CPU flags are avoided.

A future patch could make the CPU family configurable, when one wants to run
the tests on a different CPU model.

### DIFF
--- a/lago/dom_template.xml
+++ b/lago/dom_template.xml
@@ -3,7 +3,10 @@
     <memory unit='MiB'>@MEM_SIZE@</memory>
     <vcpu>@VCPU@</vcpu>
     <iothreads>1</iothreads>
-    <cpu mode='host-passthrough'>
+    <cpu match="exact">
+      <model>@CPUMODEL@</model>
+      <feature policy='optional' name='vmx'/>
+      <feature policy='optional' name='svm'/>
       <topology sockets='@CPU@' cores='1' threads='1'/>
     </cpu>
     <os>
--- a/lago/virt.py
+++ b/lago/virt.py
@@ -76,6 +76,19 @@ class VirtEnv(object):
     * libvirt_con
     '''
 
+    _CPU_FAMILIES = {
+        'Westmere': 'Intel Westmere Family',
+        'Nehalem': 'Intel Nehalem Family',
+        'Penryn': 'Intel Penryn Family',
+        'Conroe': 'Intel Conroe Family',
+        'Opteron_G5': 'AMD Opteron G5',
+        'Opteron_G4': 'AMD Opteron G4',
+        'Opteron_G3': 'AMD Opteron G3',
+        'Opteron_G2': 'AMD Opteron G2',
+        'Opteron_G1': 'AMD Opteron G1',
+    }
+    _compatible_cpu_and_family = None
+
     def __init__(self, prefix, vm_specs, net_specs):
         self.vm_types = plugins.load_plugins(
             plugins.PLUGIN_ENTRY_POINTS['vm'],
@@ -104,6 +117,18 @@ class VirtEnv(object):
         cap_tree = lxml.etree.fromstring(self.libvirt_con.getCapabilities())
         cpu_model = cap_tree.xpath('/capabilities/host/cpu/model')[0].text
         return cpu_model
+
+    def get_compatible_cpu_and_family(self):
+        if self._compatible_cpu_and_family is None:
+            for cpu in (
+                self.get_cpu_model(),
+                'Westmere',
+            ):
+                family = self._CPU_FAMILIES.get(cpu)
+                if family is not None:
+                    break
+            self._compatible_cpu_and_family = cpu, family
+        return self._compatible_cpu_and_family
 
     def _create_net(self, net_spec):
         if net_spec['type'] == 'nat':
--- a/lago/vm.py
+++ b/lago/vm.py
@@ -286,6 +286,7 @@ class LocalLibvirtVMProvider(vm.VMProviderPlugin):
             '@NAME@': self._libvirt_name(),
             '@VCPU@': self.vm._spec.get('vcpu', 2),
             '@CPU@': self.vm._spec.get('cpu', 2),
+            '@CPUMODEL@': self.vm.virt_env.get_compatible_cpu_and_family()[0],
             '@MEM_SIZE@': self.vm._spec.get('memory', 16 * 1024),
             '@QEMU_KVM@': qemu_kvm_path,
         }
--- a/ovirtlago/virt.py
+++ b/ovirtlago/virt.py
@@ -88,29 +88,7 @@ class OvirtVirtEnv(lago.virt.VirtEnv):
         return self._host_vms[:]
 
     def get_ovirt_cpu_family(self):
-        ovirt_cpu_families = {
-            'Broadwell': 'Intel Broadwell Family',
-            'Broadwell-noTSX': 'Intel Broadwell-noTSX Family',
-            'Haswell': 'Intel Haswell Family',
-            'Haswell-noTSX': 'Intel Haswell-noTSX Family',
-            'SandyBridge': 'Intel SandyBridge Family',
-            'Westmere': 'Intel Westmere Family',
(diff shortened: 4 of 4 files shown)


mykaul:
Isn't that going back to the old behavior we've had before https://github.com/lago-project/lago/pull/323 ? 
Is that a real issue, or a theoretical issue?
The first item is actually a feature of some sort - it allows to test on a more broad set of CPUs.
The last item (need update) is not a real issue - yes, you need to update the testing suite (and framework) as we (explicitly) add support to it in Engine.

Note that in Hosted Engine, this is the behavior today - we pick Conroe or so in a fixed manner.

lago/dom_template.xml
```diff
@@ -3,7 +3,9 @@
     <memory unit='MiB'>@MEM_SIZE@</memory>
     <vcpu>@VCPU@</vcpu>
     <iothreads>1</iothreads>
-    <cpu mode='host-passthrough'>
+    <cpu match="exact">
+      <model>@CPUMODEL@</model>
+      <feature policy='require' name='vmx'/>
```
mykaul: Can this work on AMD?

michalskrivanek:
> On 19 Dec 2016, at 14:48, Yaniv Kaul <notifications@github.com> wrote:
> 
> @mykaul commented on this pull request.
> 
> In lago/dom_template.xml <https://github.com/lago-project/lago/pull/398#pullrequestreview-13548905>:
> 
> > @@ -3,7 +3,9 @@
>      <memory unit='MiB'>@MEM_SIZE@</memory>
>      <vcpu>@VCPU@</vcpu>
>      <iothreads>1</iothreads>
> -    <cpu mode='host-passthrough'>
> +    <cpu match="exact">
> +      <model>@CPUMODEL@</model>
> +      <feature policy='require' name='vmx'/>
> Can this work on AMD?
> 
> 
right, we need svm on AMD, or we can add both with “optional”…that might work

michalskrivanek:
it is a real issue, we can't use pasthrough CPU as on certain models it hides some of the flags causing mis-detection in nested libvirt (namely the Lenovo W541 behves like that). We can use different CPUs but we're losing predictability...
Another option is to use libvirt cpu model with the same name as detected from the system. But I'm not sure it's going to work on all CPUs.

As for HE - well, should be the one from the cluster, ideally. But it's hard to tell when you only have the first host - you can't use the "latest" mdel as the other hosts might be older than the first one and they would be non-operational. OTOH for performance reasons it deosn't maek sense to use Conroe. Best would be if it respects the cluster, then it can be a manual procedure to go and edit the cluster CPU and restart HE VM

michalskrivanek:
Aside from host passthrough which we really should not use, I wonder what's the better approach - we get better predictability by using a single chosen CPU type (say, Westmere these days), on the other hand by using different CPU types we can expose different behavior, so worth running that too, but it still should be reproducible easily. I can imagine changing it to be configurable and in CI environment to run the tests on one of the supported types (randomly?), and if it fails re-run the tests on Conroe. Or maybe have a dedicated test for VMs using various CPU types (but we rarely have issues regarding CPUs other than on Windows...which we won't ever fully test in lago).
Anyway, for now it would be best to go with Westmere as that is a reasonable baseline, and one of the Opterons for AMD

mykaul:
On Tue, Dec 20, 2016 at 11:39 AM, michalskrivanek <notifications@github.com>
wrote:

> Aside from host passthrough which we really should not use, I wonder
> what's the better approach - we get better predictability by using a single
> chosen CPU type (say, Westmere these days), on the other hand by using
> different CPU types we can expose different behavior, so worth running that
> too, but it still should be reproducible easily. I can imagine changing it
> to be configurable and in CI environment to run the tests on one of the
> supported types (randomly?), and if it fails re-run the tests on Conroe. Or
> maybe have a dedicated test for VMs using various CPU types (but we rarely
> have issues regarding CPUs other than on Windows...which we won't ever
> fully test in lago).
>

I would actually go with Conroe in CI (just as we still have in 4.0 -
that's what the VMs will see in L2), and with whatever the host publishes
for L1, which is considered safe (I wasn't aware passthrough is not good
enough) when running outside CI (in the hope it'll be a bit faster and also
test a more interesting scenario).
I would actually think that if we can ensure CI has >Conroe across its
clusters (which I'm sure it has), it should have the highest possible
(Westmere?).

mz-pdm:
I fixed the virtualization features to provide svm as well as vmx, it works. I tried to set the (virtual) host CPU in dom_template.xml explicitly (instead of passthrough) but then the flags are still missing in the Lago hosts and the hosts are reported as non-operational by Engine for that reason. 
It would be nice to use the (virtual) host platform and make the CPU configurable. But I currently don't know how to do it in a safe manner.

michalskrivanek:
trouble is when you run lago in L1 already, then you have issues with the lago engine and hosts already seeing "different" CPU. Either way, it may very well be the different matching in kernel vs libvirt (we saw that in the past several times). It may work when the detection would use libvirt in the first place instead of /proc/cpuinfo. 
Either way, I think having Westmere with a Conroe fallback and OpteronG1 is good enough. We will not actually test anything useful until we have images with Windows where it matters.

mykaul:
On Tue, Dec 20, 2016 at 12:26 PM, michalskrivanek <notifications@github.com>
wrote:

> trouble is when you run lago in L1 already, then you have issues with the
> lago engine and hosts already seeing "different" CPU. Either way, it may
> very well be the different matching in kernel vs libvirt (we saw that in
> the past several times). It may work when the detection would use libvirt
> in the first place instead of /proc/cpuinfo.
>

But the detection does use libvirt:
        cap_tree = lxml.etree.fromstring(self.libvirt_con.getCapabilities())
        cpu_model = cap_tree.xpath('/capabilities/host/cpu/model')[0].text

michalskrivanek:
it does, but then the passthrough doesn't work correctly. Some flags are gone, and even though kernel still says it's Haswell-noTSX, libvirt identifies it as Sandy Bridge.
I would thing the performance benefits of using Haswell are negligent and not worth the side effects.

eedri:
ci test please

mykaul:
On Thu, Dec 22, 2016 at 10:55 AM, michalskrivanek <notifications@github.com>
wrote:

> it does, but then the passthrough doesn't work correctly. Some flags are
> gone, and even though kernel still says it's Haswell-noTSX, libvirt
> identifies it as Sandy Bridge.
> I would thing the performance benefits of using Haswell are negligent and
> not worth the side effects.
>

So we can map >Sandy to just being Sandy Bridge.
I agree the performance benefits are probably low.
Y.

nvgoldin:
@mz-pdm, can you rebase on master?

mz-pdm:
@nvgoldin, done.

nvgoldin:
ci test please

nvgoldin:
@mz-pdm, I added you as a member to the project so you should be able to trigger CI tests with 'ci test please'

nvgoldin:
@mz-pdm, the tests are failing on check_style, you need to format the code with ``yapf``, you can install it via pip(ensure you have version 0.14) and then run under your root lago git directory:
```
yapf --style .style.yapf --in-place --recursive .
```

mz-pdm:
ci test please

nvgoldin:
ci merge please

Issue #160, opened 2016-03-07
Add a way to export a vms disk from a prefix


david-caro:

david-caro:
this might be a step of #159

PR #401, opened 2016-12-22
Added the ability to export images from a prefix

gbenhaim:
1. Added the ability to export images from the current prefix,
   either as  a layered image or as a base image (Lago will create a
   new image which is a merge of the layer and its base).
   When exporting the image as a layer, the path to his parent will
   be replaced with the current directory, relative to the path of the
   exported image (in simple words, it expect that the parent will be with
   him in the same directory). All this logic can be found in export.py.

2. Added the ability to resolve the parent of a layered image
   (for now only one level, and only for "template_type=qcow2").
   Lago will check if the parent is in the cache, if not it will
   download it from the template repo that was provided, and then
   create a symlink from the layer's directory to the parent.

3. When using a file based disk, allow the user to choose between
   using the actual file, or let Lago make a copy of the disk.
   The default is to use the actual file, and it can be changed through
   the init file with specifying 'make_a_copy=true' under the disk
   section.

4. Added some utility functions to support the flow that was explained
   above.

5. Moved hash calculation function from templates.py to utils.py.

6. Added a verification to 'qemu-img convert' in template.py.

7. Added functional tests for the new feature.

closes https://github.com/lago-project/lago/issues/160

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -39,6 +39,7 @@ from lago.config import config
 from lago import (
     log_utils,
     workdir as lago_workdir,
+    utils,
 )
 from lago.utils import (in_prefix, with_logging)
 
@@ -272,6 +273,48 @@ def do_start(prefix, vm_names=None, **kwargs):
 @with_logging
 def do_stop(prefix, vm_names, **kwargs):
     prefix.stop(vm_names=vm_names)
+
+
+@lago.plugins.cli.cli_plugin(
+    help='Export virtual machine disks',
+    description='This command will export the disks of the given vms. '
+    'The disks of the vms will be exported to the '
+    'current directory or to the path that was specified with "-d". '
+    'If "-s" was specified, the disks will be exported as '
+    '"standalone disk", which means that they will be merged with their '
+    'base images. This command does not modifying the source disk. '
+    'The env should be in "down" state when running this command.'
+)
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--vm-names',
+    '-n',
+    help='Name of the vms to export. If no name is specified, export all '
+    'the vms in this prefix.',
+    metavar='VM_NAME',
+    nargs='*',
+)
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--standalone',
+    '-s',
+    help='If not specified, export a layered image',
+    action='store_true',
+)
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--compress',
+    '-c',
+    help='If specified, compress the exported images with xz',
+    action='store_true',
+)
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--dst-dir',
+    '-d',
+    default='.',
+    help='Dir to place the exported images in',
+)
+@in_lago_prefix
+@with_logging
+def do_export(prefix, vm_names, standalone, dst_dir, compress, **kwargs):
+    prefix.export_vms(vm_names, standalone, dst_dir, compress)
 
 
 @lago.plugins.cli.cli_plugin(
@@ -870,6 +913,10 @@ def main():
 
     try:
         cli_plugins[args.verb].do_run(args)
+    except utils.LagoUserException as e:
+        LOGGER.info(e.message)
+        LOGGER.debug(e)
+        sys.exit(2)
     except Exception:
         LOGGER.exception('Error occured, aborting')
         sys.exit(1)
--- a/lago/export.py
+++ b/lago/export.py
@@ -1,0 +1,260 @@
+from abc import ABCMeta, abstractmethod
+import logging
+import functools
+import copy
+import os
+from os import path
+import time
+import json
+import shutil
+
+from . import log_utils, utils
+
+LOGGER = logging.getLogger(__name__)
+LogTask = functools.partial(log_utils.LogTask, logger=LOGGER)
+
+
+class DiskExportManager(object):
+    """
+    ExportManager object is responsible on the export process of an image
+    from the current Lago prefix.
+
+    DiskExportManger is the base class of specific ExportManagers.
+    Each specific ExportManger is responsible on the export process of an
+    image with a specific type (e.g template, file...)
+
+    Attributes:
+        src (str): Path to the image that should be exported
+        name (str): The name of the exported disk
+        dst (str): The absolute path of the exported disk
+        disk_type (str): The type of the image e.g template, file, empty...
+        disk (dict): Disk attributes (of the disk that should be exported)
+            as found in workdir/current/virt/VM-NAME
+        exported_metadata(dict): A copy of the source disk metadata, this
+            dict should be updated with new values during the export process.
+            do_compress(bool): If true, apply compression to the exported disk.
+    """
+
+    __metaclass__ = ABCMeta
+
+    def __init__(self, dst, disk_type, disk, do_compress):
+        """
+        Args:
+            dst (str): The absolute path of the exported disk
+            disk_type (str): The type of the image
+                e.g template, file, empty...
+            disk (dict): Disk attributes as found in
+                workdir/current/virt/VM-NAME
+            do_compress(bool): If true, apply compression to the
+                exported disk.
+        """
+        self.src = path.expandvars(disk['path'])
+        self.name = path.basename(self.src)
+        self.dst = path.join(path.expandvars(dst), self.name)
+        self.dist_type = disk_type
+        self.disk = disk
+        self.exported_metadata = copy.deepcopy(disk['metadata'])
+        self.do_compress = do_compress
+
+    @staticmethod
+    def get_instance_by_type(dst, disk, do_compress, *args, **kwargs):
+        """
+        Args:
+            dst (str): The path of the new exported disk.
+                can contain env variables.
+            disk (dict): Disk attributes
+                (of the disk that should be exported) as
+                found in workdir/current/virt/VM-NAME
+            do_compress(bool): If true, apply compression to the exported disk.
+        Returns
+            An instance of a subclass of DiskExportManager which
+            matches the disk type.
+        """
+        disk_type = disk['type']
+        cls = HANDLERS.get(disk_type)
+        if not cls:
+            raise utils.LagoUserException(
+                'Export is not supported for disk type {}'.format(disk_type)
+            )
+        return cls(dst, disk_type, disk, do_compress, *args, **kwargs)
+
+    @abstractmethod
+    def export(self):
+        """
+        This method will handle the export process and should
+        implemented in each subclass.
+        """
+        raise NotImplementedError(
+            'export is an abstract method that should'
+            'be implemented in subclass'
+        )
+
+    def copy(self):
+        """
+        Copy the disk using cp in order to preserves the 'sparse'
+        structure of the file
+        """
+        with LogTask('Copying disk'):
+            utils.cp(self.src, self.dst)
+
+    def sparse(self):
+        """
+        Make the exported images more compact by removing unused space.
+        Please refer to 'virt-sparsify' for more info.
+        """
+        with LogTask('Making disk sparse'):
+            # Removed unused space from the disk
+            utils.sparse(self.dst, self.disk['format'])
+
+    def calc_sha(self, checksum):
+        """
+        Calculate the checksum of the new exported disk, write it to
+        a file, and update this managers 'exported_metadata'.
+
+        Args:
+            checksum(str): The type of the checksum
+        """
+        with LogTask('Calculating {}'.format(checksum)):
+            with open(self.dst + '.hash', 'wt') as f:
+                sha = utils.get_hash(self.dst, checksum)
+                f.write(sha)
+            self.exported_metadata[checksum] = sha
+
+    def compress(self):
+        """
+        Compress the new exported image,
+        Block size was taken from virt-builder page
+        """
+        if not self.do_compress:
+            return
+        with LogTask('Compressing disk'):
+            utils.compress(self.dst, 16777216)
+            os.unlink(self.dst)
+
+    def update_lago_metadata(self):
+        with LogTask('Updating Lago metadata'):
+            self.exported_metadata['size'] = os.stat(self.dst).st_size
+            self.exported_metadata['name'] = self.name
+            self.exported_metadata['version'] = time.strftime("%Y%m%d.0")
+
+    def write_lago_metadata(self):
+        with LogTask('Writing Lago metadata'):
+            with open(self.dst + '.metadata', 'wt') as f:
+                json.dump(self.exported_metadata, f)
+
+
+class TemplateExportManager(DiskExportManager):
+    """
+    TemplateExportManager is responsible exporting images of type template.
+
+    Attributes:
+        See superclass
+    """
+
+    def __init__(self, dst, disk_type, disk, do_compress, *args, **kwargs):
+        super(TemplateExportManager, self).__init__(
+            dst, disk_type, disk, do_compress
+        )
+        self.standalone = kwargs['standalone']
+        self.src_qemu_info = utils.get_qemu_info(self.src, backing_chain=True)
+
+    def rebase(self):
+        """
+        Change the backing-file entry of the exported disk.
+        Please refer to 'qemu-img rebase' manual for more info.
+        """
+        if self.standalone:
+            rebase_msg = 'Merging layered image with base'
+        else:
+            rebase_msg = 'Rebase'
+
+        with LogTask(rebase_msg):
+            if len(self.src_qemu_info) == 1:
+                # Base image (doesn't have predecessors)
+                return
+
+            if self.standalone:
+                # Consolidate the layers and base image
+                utils.qemu_rebase(target=self.dst, backing_file="")
+            else:
+                if len(self.src_qemu_info) > 2:
+                    raise utils.LagoUserException(
+                        'Layered export is currently supported for one '
+                        'layer only.  You can try to use Standalone export.'
+                    )
+                # Put an identifier in the metadata of the copied layer,
+                # this identifier will be used later by Lago in order
+                # to resolve and download the base image
+                parent = self.src_qemu_info[0]['backing-filename']
+                parent = './{}'.format(
+                    os.path.basename(parent.split(':', 1)[1])
+                )
+                utils.qemu_rebase(
+                    target=self.dst, backing_file=parent, safe=False
+                )
+
+    def update_lago_metadata(self):
+        super(TemplateExportManager, self).update_lago_metadata()
+
+        if self.standalone:
+            self.exported_metadata['base'] = 'None'
+        else:
+            self.exported_metadata['base'] = os.path.basename(self.src)
+
+    def export(self):
+        """
+           See DiskExportManager.export
+        """
+        with LogTask('Exporting disk {} to {}'.format(self.name, self.dst)):
+            with utils.RollbackContext() as rollback:
+                rollback.prependDefer(
+                    shutil.rmtree, self.dst, ignore_errors=True
+                )
+                self.copy()
+                self.sparse()
+                self.rebase()
+                self.calc_sha('sha1')
+                self.update_lago_metadata()
+                self.write_lago_metadata()
+                self.compress()
+                rollback.clear()
+
+
+class FileExportManager(DiskExportManager):
+    """
+    FileExportManager is responsible exporting images of type file and empty.
+
+    Attributes:
+        standalone (bool): If true, create a new image which is the result of
+            merging all the layers of src (the image that we want to export).
+        src_qemu_info (dict): Metadata on src which was generated by qemu-img.
+    """
+
+    def __init__(self, dst, disk_type, disk, do_compress, *args, **kwargs):
+        super(FileExportManager, self).__init__(
+            dst, disk_type, disk, do_compress
+        )
+
+    def export(self):
+        """
+            See DiskExportManager.export
+        """
+        with LogTask('Exporting disk {} to {}'.format(self.name, self.dst)):
+            with utils.RollbackContext() as rollback:
+                rollback.prependDefer(
+                    shutil.rmtree, self.dst, ignore_errors=True
+                )
+                self.copy()
+                self.sparse()
+                self.calc_sha('sha1')
+                self.update_lago_metadata()
+                self.write_lago_metadata()
+                self.compress()
+                rollback.clear()
+
+
+HANDLERS = {
+    'file': FileExportManager,
+    'empty': FileExportManager,
+    'template': TemplateExportManager
+}
--- a/lago/plugins/vm.py
+++ b/lago/plugins/vm.py
@@ -163,6 +163,21 @@ class VMProviderPlugin(plugins.Plugin):
         """
         pass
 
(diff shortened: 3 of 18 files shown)


lago/cmd.py
```diff
@@ -271,6 +271,39 @@ def do_stop(prefix, vm_names, **kwargs):
     prefix.stop(vm_names=vm_names)
 
 
+@lago.plugins.cli.cli_plugin(help='Export virtual entities')
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--entities-names',
+    '-e',
+    help='Name of the entities to export. If no name is specified, export all'
+    'the entities in this prefix',
+    metavar='ENTITY_NAME',
```
nvgoldin: 1. about the meaning of 'entity' - from reading this I'm not sure what it does. 
2. I think here is the place to put more docs about what this command does:
``it exports all VMs into a ..., if standalone is used ....`` etc

lago/cmd.py
```diff
@@ -271,6 +271,39 @@ def do_stop(prefix, vm_names, **kwargs):
...
+    '--standalone',
+    '-s',
+    help='If not specified, export a layered image',
+    action='store_true'
+)
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--compress',
+    '-c',
+    help='If specified, compress the exported images',
+    action='store_true'
```
nvgoldin: might be better to use 'store' and give it a default of ``True`` so it could be configurable in lago.conf
(we can't configure 'store_true' ATM)

lago/export.py
```diff
@@ -0,0 +1,239 @@
...
+        return cls(dst, disk_type, disk, do_compress, *args, **kwargs)
+
+    @abstractmethod
+    def export(self):
+        """
+        This method will handle the export process and should
+         implemented in
+        each subclass.
+        """
+        raise NotImplementedError
```
nvgoldin: better add a message here to the exception('abstract method' etc)

lago/export.py
```diff
@@ -0,0 +1,239 @@
...
+        if self.standalone:
+            # Consolidate the layers and base image
+            utils.qemu_rebase(target=self.dst, backing_file="")
+        else:
+            if len(self.src_qemu_info) > 2:
+                LOGGER.error(
+                    'Layered export is currently supported for one '
+                    'layer only.  You can try to use Standalone export.'
+                )
+                sys.exit(1)
```
nvgoldin: why exit here? better throw an exception and let it propagate(like in the rest of the code).

lago/export.py
```diff
@@ -0,0 +1,239 @@
...
+        with LogTask('Exporting disk {} to {}'.format(self.name, self.dst)):
+            with utils.RollbackContext() as rollback:
+                rollback.prependDefer(
+                    shutil.rmtree, self.dst, ignore_errors=True
+                )
+                self.copy()
+                self.sparse()
+                self.rebase()
+                self.calc_sha('sha1')
+                self.calc_sha('sha512')
```
nvgoldin: do we always need the two hashes?

lago/export.py
```diff
@@ -0,0 +1,239 @@
...
+                self.update_lago_metadata()
+                self.write_lago_metadata()
+                self.compress()
+                rollback.clear()
+
+
+class FileExportManager(DiskExportManager):
```
nvgoldin: missing docstrings

lago/export.py
```diff
@@ -0,0 +1,239 @@
...
+    def __init__(self, dst, disk_type, disk, do_compress, *args, **kwargs):
+        super(FileExportManager, self).__init__(
+            dst, disk_type, disk, do_compress
+        )
+
+    def export(self):
```
nvgoldin: docstrings

lago/plugins/vm.py
```diff
@@ -174,6 +174,29 @@ def vnc_port(self, *args, **kwargs):
         """
         pass
 
+    def export_disks(
```
nvgoldin: docstrings

lago/utils.py
```diff
@@ -561,3 +562,146 @@ def _add_subparser_to_cp(cp, section, actions, incl_unset):
         cp.set(section, var, str(action.default))
     if len(cp.items(section)) == 0:
         cp.remove_section(section)
+
+
+def with_validation(failure_msg):
```
nvgoldin: I think this decorator is problematic:
1. Use-case: it abstracts away the ability to handle the exceptions inside each function.
2. It relays on 'failure_on_error' to be defined on the decorated function, but it does not enforce setting it, so I would argue its less pythonic: explicit is better than implicit.
3. It relays that the function always returns the return code and specifically that it returns CommandStatus.

I think just updating 'run_command' with those parameters would be simpler: i.e.
```python
result = run_command(cmd=cmd, fail_on_error=True)
```

lago/utils.py
```diff
@@ -561,3 +562,146 @@ def _add_subparser_to_cp(cp, section, actions, incl_unset):
...
+        object: if backing_chain == True then a list of dicts else a dict
+    """
+
+    cmd = ['qemu-img', 'info', '--output=json', path]
+
+    if backing_chain:
+        cmd.insert(-1, '--backing-chain')
+
+    result = run_command(cmd)
+    if result and fail_on_error:
```
nvgoldin: the decorator is not used on purpose?

lago/utils.py
```diff
@@ -561,3 +562,146 @@ def _add_subparser_to_cp(cp, section, actions, incl_unset):
...
+
+
+@with_validation('Failed to run qemu-img rebase')
+def qemu_rebase(target, backing_file, safe=True):
+    """
+    changes the backing file of 'source' to 'backing_file'
+
+    Args:
+        target(str): Path to the source disk
+        backing_file(str): path to the base disk
```
nvgoldin: worth documenting what happens when ``backing_file``  is empty(as used in the code)

lago/utils.py
```diff
@@ -561,3 +562,146 @@ def _add_subparser_to_cp(cp, section, actions, incl_unset):
...
+        '--format',
+        input_format,
+        '--in-place',
+        input_file,
+    ]
+    result = run_command(cmd)
+
+    return result
+
+
```
nvgoldin: I think all of the qemu related command should be moved to a different file - maybe something like image_utils.py or making utils itself a module with 'images' inside.  ``utils.py`` is getting too big and out of context.

lago/utils.py
```diff
@@ -561,3 +562,146 @@ def _add_subparser_to_cp(cp, section, actions, incl_unset):
...
+    Args:
+        file_path (str): Path to the file to generate the hash for
+        checksum (str): hash to apply, one of the supported by hashlib, for
+            example sha1 or sha512
+
+    Returns:
+        str: hash for that file
+    """
+
+    sha = getattr(hashlib, checksum)()
```
nvgoldin: Might consider passing the hashing function directly instead of a string(safer and flexible)

lago/virt.py
```diff
@@ -166,6 +167,32 @@ def virt_path(self, *args):
     def bootstrap(self):
         utils.invoke_in_parallel(lambda vm: vm.bootstrap(), self._vms.values())
 
+    def export_entities(self, entities_names, standalone, dst_dir, compress):
```
nvgoldin: Missing docstrings

something not critical - but I think 'entities' as  a name doesn't mean anything, it would be hard to deduce from the name only what this function does.

lago/virt.py
```diff
@@ -166,6 +167,32 @@ def virt_path(self, *args):
...
+                    running_ents.append(ent)
+            except KeyError:
+                LOGGER.error('Entity {} does not exist'.format(name))
+                sys.exit(1)
+
+        if running_ents:
+            for ent in running_ents:
+                LOGGER.error('{} must be off'.format(ent.name()))
+            sys.exit(1)
+        # TODO: run the export task in parallel
```
nvgoldin: just wondering - what shared-resources prevent it ATM from running in parallel?

tests/functional/common.bash
```diff
@@ -20,6 +20,20 @@ VERBS=(
...
+          --out-format flat \
+          status
+    )
+    echo "$status" | grep -q "status: up"
+    if [[ $? -ne 0 ]]; then
+        return 0
+    else
+        return 1
+    fi
+}
```
nvgoldin: you need to add the verb in ``common.bash`` also in https://github.com/lago-project/lago/blob/master/tests/functional/common.bash#L3

tests/functional/export.bats
```diff
@@ -0,0 +1,167 @@
...
+    local workdir="$FIXTURES"/workdir
+    local suite="$FIXTURES"/suite.json
+
+    rm -rf "$WORKDIR" "$STORE"
+
+    # This is needed to be able to run inside mock, as it uses some temp files
+    # and that is not seamlesly reachable from out of the chroot by
+    # libvirt/kvm
+    export BATS_TMPDIR BATS_TEST_DIRNAME
+    export LIBGUESTFS_DEBUG=1 LIBGUESTFS_TRACE=1
```
nvgoldin: can be removed - see https://github.com/lago-project/lago/pull/391

tests/functional/export.bats
```diff
@@ -0,0 +1,167 @@
...
+    echo "$content" > "dummy_file"
+
+    helpers.run_ok "$LAGOCLI" \
+        copy-to-vm \
+        "$VM_NAME" \
+        dummy_file \
+        /root/dummy_file_inside
+
+    # Just to be sure that the file was written to disk
+    sleep 2
```
nvgoldin: as we extract files with SCP now, I think this is not needed.

tests/functional/export.bats
```diff
@@ -0,0 +1,167 @@
...
+    helpers.run_nook "$LAGOCLI" \
+        "export" \
+        --standalone \
+        --dst-dir "$STAND_ALONE_EXPORT_DIR"
+}
+
+@test "export.stop" {
+    cd "$FIXTURES"
+
+    "$LAGOCLI" shell "$VM_NAME" "cd /sbin && ./poweroff"
```
nvgoldin: why is this needed before export? and if so - it should be documented in the CLI.

tests/functional/export.bats
```diff
@@ -0,0 +1,167 @@
...
+    sleep 5
```
nvgoldin: can it be less than 5?

tests/functional/export.bats
```diff
@@ -0,0 +1,167 @@
...
+    cd "$FIXTURES"
+
+    helpers.run_ok common.is_stopped $WORKDIR
+    helpers.run_ok "$LAGOCLI" \
+        "export" \
+        --standalone \
+        --dst-dir "$STAND_ALONE_EXPORT_DIR"
+}
+
+@test "export.export_layred: layred images export" {
```
nvgoldin: s/layred/layered

tests/functional/export.bats
```diff
@@ -0,0 +1,167 @@
...
+        "$suite"
+}
+
+@test "export.test_exported_env: testing standalone export" {
+    cd $FIXTURES
+    local dummy_file_content=$(cat dummy_file)
+    cd "$STAND_ALONE_EXPORT_DIR"
+
+    helpers.run_ok "$LAGOCLI" start
+    dummy_file_inside_content=$("$LAGOCLI" shell "$VM_NAME" "cat /root/dummy_file_inside")
```
nvgoldin: not directly related - but just emphasises the fact we need to fix https://github.com/lago-project/lago/issues/409 as all of these sort of checks might fail too.
btw one way to overcome this is to generate the dummy file with N lines and compare only its 'tail -N'

nvgoldin:
Great job!
2 more things:
1. Might be worth to add to the PR description few running commands examples(and they can later be added to the docs), so anyone who wants(like me:) to test it would know how to run it easily. 

2. Major issue I have is with the tests, currently it can take up to 700 seconds to run the functional tests in CI, I think we cannot afford to increase the running time any more, it makes developement and release process a pain. So first question is how long does it take?
second - I think you should merge the standalone export tests into other existing tests, you already have everything in there(the dummy file, etc) so we could skip the expensive operation of running 'init' again. About the layered images test - need to check how much time does it take, this might also be possible to merge into the existing suite(needs more checking though)

lago/prefix.py
```diff
@@ -493,7 +493,17 @@ def _handle_file_disk(self, disk_spec):
         if url:
             disk_spec['path'] = self._retrieve_disk_url(url, disk_path)
         else:
-            disk_spec['path'] = disk_path
+            # Create a copy of the file or use the original
+            if disk_spec.get('make_a_copy'):
+                LOGGER.info("Making a copy")
+                dest_path = self._generate_disk_path(
+                    os.path.basename(disk_spec['path'])
+                )
```
nvgoldin: not sure - would we have a problem here with >1 disks from the same image and ``'make_a_copy=true'``?

lago/export.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+        raise NotImplementedError
+
+    def copy(self):
+        """
+        Copy the disk using cp in order to preserves the 'sparse'
+        structure of the file
+        """
+        with LogTask('Copying disk'):
+            utils.cp(self.src, self.dst)
```
mykaul: Does that copy with sparseness? If not (and even if it does), we should probably run 'virt-sparsify' on the image first. It takes few seconds, and reduces nicely (by up to several hundreds of megs) our disks. virt-sparsify is better than plain 'cp' with sparseness preserveness, as it 'compacts' the image as well - sparsifies marked-as-deleted-but-not-deleted blocks.

lago/export.py
gbenhaim: I think that printing a huge stack trace to the screen is a bad user user experience, and should be used only when an actual exception was occurred. IMO when dealing with a known issue it's better to print an explanation on the error and suggest a way to fix it.

lago/export.py
gbenhaim: I followed the convention that is being used in Lago Images, even though we don't really use 'sha512'. The verification of the image is done with  'sha128'. If we decide to drop 'sha512' in the export feature we also need to drop it from Lago Images (I don't mind to drop it). 
What do you think ?

tests/functional/export.bats
gbenhaim: Clean shutdown isn't needed before exporting, you can just stop the env using 'lago stop'.
In the case of the functional test, if not shutting down the machine properly the 'dummy_file_insdie' isn't written to disk.

lago/prefix.py
gbenhaim: Can you give an example ?

nvgoldin: something like:
```yaml
domains:
  dom1:
..
    disks:
      - template_name: disk1
        path: '$PWD/disk1.qcow2'
        make_a_copy: True
        type: file
        ....

  dom2:
..
    disks:
      - template_name: disk1
        path: '$PWD/disk1.qcow2'
        make_a_copy: True
        type: file
        ....
```

Or even:
```yaml
domains:
  dom1:
..
    disks:
      - template_name: disk1
        path: '$PWD/disk1.qcow2'
        make_a_copy: True
        type: file
        ....
    - template_name: disk2
        path: '$PWD/disk1.qcow2'
        make_a_copy: True
        type: file
....
```

The idea is to check that the copy we make has a unique name.

lago/export.py
nvgoldin: I think if its functional with just one of them, we should keep that way and open an issue to order the to decide what to use. 
If for some reason its not functional with both, even though we just use sha128, then we need to fix that. Maybe just add some dummy hash for now? (i.e. 'not_used_yet). As calculating the hash is relatively expensive.

lago/export.py
nvgoldin: I agree about the user-experience, but I don't think this is easily solved by ``sys.exit``:
1. The rest of the code(I think) doesn't do that, which will cause inconsistencies when trying to debug issues. 
2. The 'rebase' command is not a direct CLI command, other functions may call it in different scenarios, not even necessarily as in this PR, so there will be no way to know if they are responsible for a wrong call, or it exits as expected here. 
3. We might have right now(didn't check) or have in the future cleanup code that should be called before the program terminates. Simply exiting the program abruptly might prevent it from running (python raises ``SystemExit`` exception on ``sys.exit`` so I think 'finally' blocks up the stracktrace should be executed, but other stuff might not)
4. Following 3- I'm not sure how python will behave as far as memory cleanup and object releases when issued sys.exit(1), I would guess right now its not a big issue, but might be.
5. There is no logging of the actual exception and its stacktrace. 

If you still want to change this user behaviour, an alternative that comes to mind:

Add in the export function(the direct call the user makes) in  ``cmd.py`` something like:
```python
try:
    ... do your stuff here
catch (your expected exception list) as e:
   LOGGER.info(e.message)
   LOGGER.exception(e)
   sys.exit(1)
```
I can't say I'm 100% pleased with that, as we still swallow the exception(or more correctly raise a different one - ``SystemExit``, instead of the original one), but at least it will be logged to ```lago.log``` and done as close as possible 'up' to the stacktrace, so whoever wanted to clean-up his mess had a chance to do so. 


Further ahead: we can do the same technique for all exception, maybe add a flag that will control whether to print the stacktrace, if set to false, print the message to stdout, and log the entire exception to lago.log. Alternatively, we can control the depth of the printed stacktrace with ```traceback.print_tb```

tests/functional/export.bats
nvgoldin: 👍 
Could also add a check before running export that verifies the VMs are down.

lago/utils.py
gbenhaim: What do you mean ? Can you give an example ?

nvgoldin: change:
```python
def get_hash(file_path, checksum='sha1'):
```
to:
```
def get_hash(file_path, checksum=hashlib.sha1):
```
So if the caller wants a different function, he would pass it directly, so you don't need to do 'string' -> function conversions.

tests/functional/export.bats
gbenhaim: It exists in virt.py, also added a functional test to check it.

lago/virt.py
gbenhaim: Not something that I can think on for this moment, but let keep things as simple as we can for now, and improve later.

tests/functional/export.bats
gbenhaim: I don't see how it relates, I'm using 'copy-to-vm' which always used ssh, am I missing something?

lago/prefix.py
gbenhaim: I can't see a situation where one disk is being used by two vms, nor used twice or more times by one vm.

lago/export.py
gbenhaim: This copy method is based the os' "cp", which indeed preserve the sparseness  of the file. After the copy I'm running virt-sparsify in "--in-place" mode, which doesn't need large amounts of temporary space.

lago/export.py
gbenhaim: for the meanwhile I removed sha512 calculation, as it takes a lot of time and not really being used.

tests/functional/export.bats
nvgoldin: I meant the ``sleep 2``, as we user SCP to extract the files, it should have no affect.

tests/functional/export.bats
nvgoldin: 👍

lago/virt.py
nvgoldin: 👍

lago/export.py
```diff
@@ -203,20 +215,31 @@ def export(self):
...
 
 
 class FileExportManager(DiskExportManager):
+    """
+    FileExportManager is responsible exporting images of type file and empty.
+
+    Attributes:
+        standalone (bool): If true, create a new image which is the result of
+         merging all the layers of src (the image that we want to export).
+         src_qemu_info (dict): Metadata on src which was generated by qemu-img.
```
nvgoldin: I think you need to keep the same indentation, otherwise it would not format nicely with sphinix:
```python
"""
Attributes:
    standalone (bool): long line ...
        more of that long line
    othervar (str): blah
"""
```

lago/utils.py
```diff
@@ -564,37 +564,14 @@ def _add_subparser_to_cp(cp, section, actions, incl_unset):
...
-                raise RuntimeError(
-                    'msg: {}\nfunction: {}\nsubprocess msg: '
-                    '"{}"\nargs: {}\nkwargs: {}'.format(
-                        failure_msg, func.__name__, result.err, args, kwargs
-                    )
-                )
-            return result
-
-        return wrapped
+def run_command_with_validation(
```
nvgoldin: nice!

lago/utils.py
```diff
@@ -614,62 +591,63 @@ def get_qemu_info(path, backing_chain=False, fail_on_error=True):
...
 
-@with_validation('Failed to run qemu-img rebase')
-def qemu_rebase(target, backing_file, safe=True):
+def qemu_rebase(target, backing_file, safe=True, fail_on_error=True):
     """
     changes the backing file of 'source' to 'backing_file'
+    If backing_file is specified as "" (the empty string),
+    then the image is rebased onto no backing file
+    (i.e. it will exist independently of any backing file).
+    (Taken from qemu-img man page)
```
nvgoldin: thanks for adding that, makes things clear.

lago/virt.py
```diff
@@ -192,31 +191,33 @@ def virt_path(self, *args):
...
+                )
 
-        if running_ents:
-            for ent in running_ents:
-                LOGGER.error('{} must be off'.format(ent.name()))
-            sys.exit(1)
+        if running_vms:
+            raise utils.LagoUserException(
+                'The following vms must be off:\n{}'.
+                format('\n'.join([_vm.name() for _vm in running_vms]))
```
nvgoldin: nice

lago/virt.py
```diff
@@ -192,31 +191,33 @@ def virt_path(self, *args):
...
+            )
         # TODO: run the export task in parallel
 
         with LogTask('Exporting disks to: {}'.format(dst_dir)):
-            for ent in entities:
-                ent.export_disks(standalone, dst_dir, compress)
+            for _vm in vms:
```
nvgoldin: just wondering - why ``_vm``? implies that is a private variable

tests/functional/export.bats
```diff
@@ -132,7 +131,7 @@ export LIBGUESTFS_BACKEND=direct
     cd "$STAND_ALONE_EXPORT_DIR"
 
     helpers.run_ok "$LAGOCLI" start
-    dummy_file_inside_content=$("$LAGOCLI" shell "$VM_NAME" "cat /root/dummy_file_inside")
+    dummy_file_inside_content=$("$LAGOCLI" shell "$VM_NAME" "cat /root/dummy_file_inside" | tail -n1)
```
nvgoldin: cool - though that file is indeed with more than one line? (otherwise it still might fail)

lago/virt.py
gbenhaim: Otherwise the formatter is complaining about the re-assignment of varaible "vm" . I just needed another name for the variable.

gbenhaim:
ci merge please

nvgoldin:
ci test please

gbenhaim:
ci test please

gbenhaim:
ci merge please

PR #405, opened 2016-12-25
Check network is active after creation

nvgoldin:
This will busy-wait until the network is created
for at most 10 seconds. At the minimum, it will
exit after the first check the network is active,
without sleeping.

### DIFF
--- a/lago/virt.py
+++ b/lago/virt.py
@@ -23,7 +23,7 @@ import json
 import logging
 import os
 import uuid
-
+import time
 import lxml.etree
 
 from . import (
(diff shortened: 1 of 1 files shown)


nvgoldin:
ci test please

lago/virt.py
```diff
@@ -348,14 +348,50 @@ def resolve(self, name):
...
+                if net is None:
+                    raise RuntimeError(
+                        'failed to create network, XML: %s' %
+                        (self._libvirt_xml())
+                    )
+                for _ in range(attempts):
+                    if net.isActive():
+                        return
+                    LOGGER.debug(
+                        'waiting for network %s to become active', net.name()
```
mykaul: Is it self.name() or net.name() ? same below.

nvgoldin: took it directly from libvirt instead(we already know its not ``None`` here..)

nvgoldin:
ci merge please

Issue #368, opened 2016-11-27
'lago ovirt reposetup' fails if the reposync config file does not exist


nvgoldin:
Even though it is still functional with 'custom_source' only.

PR #408, opened 2016-12-26
ovirtlago: allow running reposync with custom_sources only

nvgoldin:
This fixes a small bug, when the 'reposync' configuration
was passed but didn't exist, the 'ovirt reposetup' command
failed.

fixes: https://github.com/lago-project/lago/issues/368

### DIFF
--- a/ovirtlago/__init__.py
+++ b/ovirtlago/__init__.py
@@ -103,7 +103,6 @@ class OvirtPrefix(Prefix):
         skip_sync=False,
         custom_sources=None
     ):
-        custom_sources = custom_sources or []
         # Detect distros from template metadata
         engine_dists = [self.virt_env.engine_vm().distro()] \
             if self.virt_env.engine_vm() else []
@@ -114,7 +113,7 @@ class OvirtPrefix(Prefix):
 
         repos = []
 
-        if rpm_repo and reposync_yum_config:
+        if rpm_repo is not None and reposync_yum_config is not None:
             parser = SafeConfigParser()
             with open(reposync_yum_config) as repo_conf_fd:
                 parser.readfp(repo_conf_fd)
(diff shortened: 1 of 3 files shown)


ovirtlago/__init__.py
```diff
@@ -114,7 +114,7 @@ def prepare_repo(
 
         repos = []
 
-        if rpm_repo and reposync_yum_config:
+        if rpm_repo and os.path.isfile(reposync_yum_config):
```
ifireball: wouldn't it be cleaner to just ensure that reposync_yum_config gets initialized to None instead of a non-existent file name?
I think we do want to raise an error if the used specified a non-existent file on the command line...

nvgoldin:
@ifireball  - ping :)

ovirtlago/cmd.py
```diff
@@ -35,15 +35,6 @@
     workdir_class=ovirtlago.OvirtWorkdir,
 )
 # TODO: Remove this, and properly complain on unset config values
```
ifireball: you probably just did that

ovirtlago/__init__.py
```diff
@@ -133,6 +133,17 @@ def prepare_repo(
                         reposync_yum_config,
                         repos,
                     )
+        elif custom_sources:
```
ifireball: its better style to use 'is [not] Null' rather then the truthiness check. It also better UX IMO, if someone passes an empty string for the argument on the command line, he should get a "file not found" error, not the default behaviour...

ifireball: this comment also applies to the above reposync_yum_config check (not patched here)

ovirtlago/cmd.py
nvgoldin: nope, theres still
```
DISTS = ['el6', 'el7', 'fc20']
```
which I will remove in a next clean-up patch(there many other stuff here too..)

nvgoldin:
@ifireball  - done.

nvgoldin:
ci merge please

nvgoldin:
ci merge please

PR #415, opened 2017-01-04
ovirtlago: collect entire /var/log directory

nvgoldin:
Collect entire '/var/log' directory, and instead of "fixing" the ovirt.collect functional test, just dropping them, as 'lago ovirt collect' is deprecated.


landgraf:
Isn't it better to collect entire /var/log directory? audit.log can be useful in some cases as well as secure.log

nvgoldin:
Theoretically I think its ok, but:
Current uncompressed size of OST logs for each run is around 90MB .

Maybe adding what that is left in /var/log is not a lot, need to check. If it raises the size significantly i think we rather pre-define what we need.

mykaul:
On Wed, Jan 4, 2017 at 12:30 PM, Nadav Goldin <notifications@github.com>
wrote:

> Theoretically I think its ok, but:
> Current uncompressed size of OST logs for each run is around 90MB .
>

But compressed it's probably far less?

> Maybe adding what that is left in /var/log is not a lot, need to check. If
> it raises the size significantly i think we rather pre-define what we need.
>

I doubt it matters. We need to reduce image sizes by some hundreds -
that'll make a difference ;-)

nvgoldin:
done - in engine/host/HE  VMs the entire /var/logs will be collected. Uncompressed size is ~100MB.

nvgoldin:
ci merge please

ifireball:
We should at least make OST cleaup the /var dir before any test so that you get results that are relevant only for that test. Probably a lot of that 100m is old stuff we don't need.

nvgoldin:
>We should at least make OST cleaup the /var dir before any test so that you get results that are relevant only for that test. Probably a lot of that 100m is old stuff we don't need

Yes but I think that would make it hard to debug, you'll have for instance 'engine.log' split into 4 different directories, and things might be dependant. Also while you 'mv/rm' the directory, stuff might get written too.

I suggest we try extracting the entire '/var/log' and if it becomes a problem in any way, we can always filter again.

ifireball:
At the very least we need to to it before the deployment, to not carry around old stuff...

nvgoldin:
> At the very least we need to to it before the deployment, to not carry around old stuff...

I don't think it takes a lot of space, just the initial boot-up logs.

About rotating the logs, this will require some more re-factoring, so I don't think its for the scope of this patch.

ifireball:
I didn`t say you need to rotate the logs, just clean up rotated files.

This isn't about space, this is about reducing the haystack you need to look at when debugging issues.

nvgoldin:
Ok, opened issue in: https://github.com/lago-project/lago/issues/420

nvgoldin:
ci merge please

PR #418, opened 2017-01-05
Force usage of qemu-kvm-ev on centos/rhel

gbenhaim:
Lago requires qemu-kvm >= 2.1.2 and on centos/rhel those versions
are provided by qemu-kvm-ev.

### DIFF
--- a/automation/check-merged.repos.el7
+++ b/automation/check-merged.repos.el7
@@ -1,0 +1,1 @@
+check-patch.repos.el7
--- a/automation/check-patch.repos.el7
+++ b/automation/check-patch.repos.el7
@@ -1,0 +1,2 @@
+ovirt-ci-tools,http://resources.ovirt.org/repos/ci-tools/$distro
+centos-qemu-ev,http://mirror.centos.org/centos/7/virt/x86_64/kvm-common/
--- a/docs/Installation.rst
+++ b/docs/Installation.rst
@@ -19,7 +19,8 @@ For Fedora::
   enabled=1
   gpgcheck=0
 
-For EL distros (such as CentOS, RHEL, etc.), make sure you have EPEL and::
+For EL distros (such as CentOS, RHEL, etc.), make sure you have epel-release
+and centos-release-qemu-ev repositories installed and enabled, and::
 
   [lago]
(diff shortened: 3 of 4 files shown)


gbenhaim:
@rafaelmartins

gbenhaim:
ci merge please

nvgoldin:
@rafaelmartins - I think there is an issue here with el(not centos) - the package is called 'qemu-kvm-rhev', no?

gbenhaim:
@rafaelmartins,
@nvgoldin is right. Is there any way to distinguish between rhel and cetnos in the rpm spec?

nvgoldin:
We could do something like this:
```
+
+%if 0%{?rhel} && 0%{?centos}
+Requires: qemu-img-ev >= 2.1.2
+Requires: qemu-kvm-ev >= 2.1.2
+%endif
+
+%if 0%{?rhel} && ! 0%{?centos}
+Requires: qemu-img-rhev >= 2.1.2
+Requires: qemu-kvm-rhev >= 2.1.2
+%endif
+
+%if 0%{?fedora}
 Requires: qemu-img >= 2.1.2
 Requires: qemu-kvm >= 2.1.2
+%endif

```
But, I'm not sure if this is a problem, can we just install ``qemu-kvm-ev`` on rhel? @sandrobonazzola - is that compatible?

rafaelmartins:
Looking at vdsm spec file, it seems that we can only require -rhev, and it will work for centos. -ev probably provides -rhev.

https://gerrit.ovirt.org/gitweb?p=vdsm.git;a=blob;f=vdsm.spec.in;h=0d658817b99d69fb83f10b2c031db896fe478fb9;hb=HEAD#l186

nvgoldin:
👍  - just tested on a fresh CentOS 7 and ``qemu-kvm-rhev`` works - could you update the PR?

rafaelmartins:
I don't think I have access, it must be you or @gbenhaim, probably :)

nvgoldin:
ci test please

nvgoldin:
@rafaelmartins  - yep not sure why this was created on master :) 
anyways - updated the PR, thanks.

nvgoldin:
ci merge please

Issue #399, opened 2016-12-21
functional tests are failing on 'shell to vm' 


nvgoldin:
The CirrOS VM doesn't get an IP, logging in with lago console and executing ```/etc/init.d/S40network restart``` resolves the issue. The only change done to automation was https://github.com/lago-project/lago/commit/afec1ac3042adb62678e0940fd2289226e9aa129, could it be that ``dnsmasq`` is not yet configured when we run ``lago start``(now with KVM enabled it is probably done faster than before)? @david-caro - any ideas?

mykaul:
I would say that this is exactly why we need to make some order between SSH and root passwords, between password-based and public-key based authentication. We need to see that such a VM:
1. Does have a root password (for ovirt-system-tests use case)
2. Does have a SSH password (which happens of course to be the root password) - for Lago to connect to it.
3. Does not have a SSH public key (we do not inject to it a public key - note that ovirt-system-tests might in the future via cloud-init

Currently, there's quite a bit of a mixup. It's very similar to the situation with Hosted-Engine.

gbenhaim:
We've already experienced this issue before: https://github.com/lago-project/lago/pull/318
The interface is shown in the vm, but for some reason ifup/ifdown can't recognize it.  I don't think that this issue relates to ssh but also don't have a good explanation why it relates to the boot menu.

mykaul:
On Thu, Dec 22, 2016 at 5:50 PM, Gal Ben Haim <notifications@github.com>
wrote:

> We've already experienced this issue before: #318
> <https://github.com/lago-project/lago/pull/318>
> The interface is shown in the vm, but for some reason ifup/ifdown can't
> recognize it. I don't think that this issue relates to ssh but also don't
> have a good explanation why it relates to the boot menu.
>

Reminds me of an infamous issue in OpenStack - the network (and
specifically, the DHCP services) are not fast enough to be up and running
before the VM is already requesting the network.
Y.

gbenhaim:
As I see it we have two options:
1. Enable the boot menu.
2. Search for another image to use in the functional tests.

@mykaul @nvgoldin @ifireball @david-caro 
Thoughts ?

mykaul:
I've opened a bug on libvirt on the issue.
Y.

nvgoldin:
I looked at dnsmasq logs:

```
[...]

PR #419, opened 2017-01-05
tests: update cirros image to start cirros-dhcpc explicitly

nvgoldin:
This should resolve the problem that the image didn't get an IP during the
tests, if it boot up too 'quickly', which might be caused by
a bug in dnsmasq/libvirt(the configurations were not deployed yet).

issue: https://github.com/lago-project/lago/issues/399


ifireball:
I'd rather we have a logo-level solution for this, so using fast images with lago does not become a "crime"

nvgoldin:
1. An attempt to resolve it "inside lago" was done in https://github.com/lago-project/lago/pull/405, turns out it wasn't enough.
2. We don't know yet the true core of the problem, and resolving this will require some more time(as you see I didn't add  'closes:', so the ticket is going to be left open)
3. Meanwhile, this blocks CI and causes annoying failures that add a lot of headache. And most importantly, the only place we see this happening is on the functional tests using the cirros image.

So yes, I agree, but I rather get this going now. The ticket will be left open.

mykaul:
Do we need an updated image in ovirt.org Glance server as well?

ifireball:
maybe we can `dhcping` dnsmasq? or perhaps just try a dns query against it? I also have a piece of pure Python code that can do DHCP querying, if you don't want to depend on another a command-line tool...

ifireball:
> Do we need an updated image in ovirt.org Glance server as well?

I wouldn`t want to host tweaked images there, I need to able to grantee users that images there were not tampered with.

nvgoldin:
> maybe we can dhcping dnsmasq? or perhaps just try a dns query against it? I also have a piece of pure Python code that can do DHCP querying, if you don't want to depend on another a command-line tool...

great 👍 , would be nice if you can put it somewhere and I could try to query dnsmasq before the run stage. Though as I said, I don't think it contradicts this PR. The cirros image serves us for testing, as long it does allow us to test what we want(and how quickly libvirt deploys dnsmasq configuration is not one of our tests), I don't mind tweaking it.

nvgoldin:
I ran the 'check-merged' twice on one of the hyper visors(where it happened before as supposedly it deployed the image too quickly) and all went good. 
http://jenkins.ovirt.org/job/lago_master_check-merged-el7-x86_64/303/
http://jenkins.ovirt.org/job/lago_master_check-merged-el7-x86_64/302/

nvgoldin:
I'm abandoning this in favour of https://github.com/lago-project/lago/pull/421,
1. I think https://github.com/lago-project/lago/pull/421 is better as it will allow(if needed) to increase the wait time, and the workaround is inside lago. 
2. Major issue with this PR which was brought to my attention by @ifireball: it will increase (most likely forever) our git repository by around 12MB.

mykaul:
I still have no doubt that we need to try with newer cirros images (true, they have not released one in ages, but they are available). Examples of potentially relevant commits:
https://git.launchpad.net/cirros/commit/?id=4285160b0d0b7f9be19f7464b61df3ad1decd3cb - IPv6 information
https://git.launchpad.net/cirros/commit/?id=6752f081363a0d38b1dd41593b47ac60b0caab52 - virtio-SCSI (which we need!)
 https://git.launchpad.net/cirros/commit/?id=c475073deef2e45ffc61d469a0b6254a365a6d2c - another fix in DHCP ...
https://git.launchpad.net/cirros/commit/?id=2d3e303f5f940c411feaa74917988f8c397a2ef4 - much newer kernel

nvgoldin:
@mykaul  - I agree, I'll open an issue there(when launchpad lets me register!) to check when they are going to do another release. When they do - we can test it thoroughly and pay the price of 12MB in the git repo(rather pay it once..)

mykaul:
@nvgoldin try http://download.cirros-cloud.net/daily/20161201/

PR #421, opened 2017-01-08
workaround for cirros images not getting ip on boot

nvgoldin:
When 'LAGO__START__WAIT_SUSPEND' environment variable value
is set to a float or integer, each domain will be started
in 'paused' mode(CPU halted), and it will be resumed
after waiting for the value set.

This is currently not added to the CLI, as hopefully
this workaround will be removed once the BZ is fixed.

bug-url: https://bugzilla.redhat.com/show_bug.cgi?id=1411025

### DIFF
--- a/lago/vm.py
+++ b/lago/vm.py
@@ -18,6 +18,7 @@
 # Refer to the README and COPYING files for full details of the license
 #
 import functools
+import time
 import guestfs
 import libvirt
 import logging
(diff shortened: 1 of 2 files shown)


nvgoldin:
verified locally and in http://jenkins.ovirt.org/job/lago_master_check-merged-el7-x86_64/306/console(on a hyper-visor).
Currently the wait time is set to 1 second, which seems to be enough. If we see more failures such as https://github.com/lago-project/lago/issues/399, we can increase the wait time.

nvgoldin:
Because that doesn't work :) 
The issue is not the time between the network is created and the domain, but when the domain(with its own interface) is created. I added some more details about it in the BZ.
Empirically, waiting between the network creation and the domain is created does not resolve this.

mykaul:
Doesn't it makes sense to revert the boot menu delay elimination and as well as the grub timeout=0? much simpler.

ifireball:
@mykaul why slow down OST to resolve a Lago integration testing issue?

mykaul:
On Jan 8, 2017 10:32 AM, "Barak Korren" <notifications@github.com> wrote:

@mykaul <https://github.com/mykaul> why slow down OST to resolve a Lago
integration testing issue?


For the same reason no one is testing a newer version of cirros?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/lago-project/lago/pull/421#issuecomment-271138003>, or mute
the thread

nvgoldin:
>Doesn't it makes sense to revert the boot menu delay elimination and as
well as the grub timeout=0? much simpler.

As we don't really need the bootmenu, and it will slow down runs for all
VMs, I think the added complexity is better(and again - this is just when
the variable is set, so on the common case nothing is changed).


> We might consider specifying the delay in the image metadate if the
libvirt issue remains unresolved. (Consider an environment with mix of
CentOS and CirrOS VMs, no need to make CentOS wait because of CirruOS...)

I hope this will be resolved(by libvirt or by discovering a problem in our
code or switching image) before we standardize it further :)

nvgoldin:
I know there are mixed opinions here, but as we still have the ticket to update the cirros image open(which is not contradicting and is important on its on), I'm merging this as I don't want any more CI failures.
Hopefully we will come up with a solution soon which would allow to remove this.

nvgoldin:
ci merge please

nvgoldin:
failed on a different issue(lago ovirt collect tests, which I hope to drop soon).

nvgoldin:
ci merge please

PR #424, opened 2017-01-10
Return SandyBridge CPU family

lveyde:


gbenhaim:
ci test please

nvgoldin:
I can't verify this right now, though what I suggest:
1. Test this with OST HE(needs to be done on the CI servers to make sure it works), and also @mz-pdm will test it solves his issue.
2. Revert https://github.com/lago-project/lago/commit/bb846b43286b2a562c0592b46429119d179eca68, which will break things again for @mz-pdm, but we could fix them properly later(or most likely make the CPU model configurable).

As lago's CI is pretty stable right now, making another release just to fix this issue shouldn't take long.

mz-pdm:
I tested the patch on both my laptop and our lab server and it works.

> Revert bb846b4, which will break things again for @mz-pdm, but we could fix them properly later(or most likely make the CPU model configurable).

We are trying to use Lago. But to make it feasible Lago should run out of the box. It may need tweaking for special purposes or on CI but if a developer installs Lago + ovirt-system-tests on any of his computers, it should work without any special configuration (as long as common requirements such as amount of RAM and disk space are satisfied). Please keep that on mind whatever you decide to do.

nvgoldin:
@mz-pdm, we agree here, I missed in my comment an 'or' between 1 and 2. Either way I think we need to fix both problems(in CI and on the mentioned lenovo laptop model). If this PR solves also the HE engine its fine with me.

nvgoldin:
@lveyde  - were you able to verify it on HE suite?

nvgoldin:
The CPU detected for host0:
```xml
  <cpu mode='custom' match='exact'>
    <model fallback='allow'>Westmere</model>
    <topology sockets='2' cores='1' threads='1'/>
    <feature policy='optional' name='vmx'/>
    <feature policy='optional' name='svm'/>
  </cpu>
```

nvgoldin:
For the record the server I tested this on was ``Haswell-noTSX``, as we dropped it from the CPU families mapping in https://github.com/lago-project/lago/pull/398, it detected it as the default - ``Westmere``, which conflicted with HE answer-file(``SandyBridge``). @lveyde  - that explains why it worked for you and failed for me ;)
As the hard-coded demand is actually on HE side - I think https://github.com/lago-project/lago/pull/427 should solve it, leaving no places for unknowns. On the other hand, it doesn't contradict this PR(SandyBridge -> SandyBridge), so we can get both in.

nvgoldin:
ci merge please

PR #427, opened 2017-01-15
Allow to define custom cpu model

nvgoldin:
This PR adds a 'cpu_model' parameter to the init file. If defined, it will generate the domain XML with that cpu_model, otherwise it will use the environment defaults. 
There will be 2 options to define the cpu model:
1. By defining it in the init file as noted above. 
2. By creating a new VM plugin and overriding the ``cpu_model`` property. This could allow to create more complex CPU model detections if needed.


nvgoldin:
Want to get this going today(to make OST HE suite work again) - so reviews are welcome.

nvgoldin:
ci merge please

PR #431, opened 2017-01-17
Revert "ovirtlago: drop get_ovirt_cpu_family"

nvgoldin:
This reverts commit f31b74c44e7c1d3db61c85af3cfc59846c013751.

My mistake - I missed that we call this method in OST. 
We should  probably deprecate this method and instead call the prefix direct method, but this can be done later.


nvgoldin:
ci merge please

PR #438, opened 2017-01-25
Add a get method to retrieve the SSH public key contents.

mykaul:
This will be used later to install hosts using public key, for Ansible, etc.

### DIFF
--- a/lago/prefix.py
+++ b/lago/prefix.py
@@ -1040,6 +1040,19 @@ class Prefix(object):
         """
         return self.virt_env.get_nets()
 
+    def get_ssh_pub_key(self):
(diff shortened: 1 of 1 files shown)


mykaul:
Not good enough. I need the fingerprint retrieved as well. ARGH. Don't merge yet.

mykaul:
Might be good enough (so I can later add in ovirt-system-tests a test that adds a User with SSH Public Key). If there are no comments, I'd be happy to get it in.
(For host-based SSH PK installation, I'll need the fingerprint. This will need to be worked on differently - I assume running some command on the Engine).

PR #446, opened 2017-02-07
safely check qemu uid after snapshot

nvgoldin:
In situations where lago runs under a container(or mock),
we might bind-mount the images directory into the container,
and then the images might be in an unknown uid/gid, therefore it
is not safe to compare their uid's against the environment users
database. This patch ensures the check is safe. Note that if
the caller does not have sufficient privileges, it will fail(as needed).

closes: https://github.com/lago-project/lago/issues/447


nvgoldin:
ci merge please

PR #448, opened 2017-02-08
Add Skylake to supported CPU families.

lenoch:
This makes `ovirt-system-tests` work; in particular the tests `vm_run`, `snapshots_merge`, and `suspend_resume_vm` from `004_basic_sanity`.

It seems that Westmere might not be the best fallback.


mykaul:
That's only relevant for Skylake based hosts? Why isn't Westmere working? What would customers do? Is that because of nested? What about host-passthrough?

nvgoldin:
TL;DR, I think we should go back to host-passthrough by default(as it was before https://github.com/lago-project/lago/pull/398) and extend 'cpu_model' to allow more configurations. 

1. libvirt doesn't care about 'cpu family', afaik it only serves us in OST to determine the cluster family. So first, I think the entire CPU Family detection should go back to ``ovirtlago/prefix.py`` and dropped entirely from ``virt.py``.
2. I think a better approach would be to use host-passthrough by default, BUT, allow the custom 'cpu_model' to override it. In the CPU family detection under ``ovirtlago/virt.py`` it would first check if 'cpu_model' was configured explicitly for the host, and if so, it would report its matching CPU family(from the entire list of supported cluster cpu families). Otherwise, it would use host-passthrough.

This will allow us to:
1. Overcome the issue of some (mainly laptop) CPUs not working well under the current configuration. So anyone who wants to run it differently, will simply append a different 'cpu_model' parameter to the init file.
2. If we would want to test different scenearios in CI we would just state explicitly in the init file what we target.
3. And most importantly, in the common case, host-passthrough will just work. 

One issue is that 'cpu_model' will need to be extended to support feature elements addition(when using ``exact`` mode).

lenoch:
@mykaul This is actually a follow-up to the `OST: vm_run fails for me (basic-suite-master)` thread on [ovirt-devel] and [lago-devel], quoting the error:
```
Feb 07 06:33:23 lago-basic-suite-master-host0 libvirtd[12888]: Unable to read from monitor: Connection reset by peer
Feb 07 06:33:23 lago-basic-suite-master-host0 libvirtd[12888]: internal error: qemu unexpectedly closed the monitor: 2017-02-07T11:33:23.058571Z qemu-kvm: warning: CPU(s) not present in any NUMA nodes: 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
                                                               2017-02-07T11:33:23.058826Z qemu-kvm: warning: All CPU(s) up to maxcpus should be described in NUMA config
                                                               qemu-kvm: /builddir/build/BUILD/qemu-2.6.0/target-i386/kvm.c:1736: kvm_put_msrs: Assertion `ret == n' failed.
```

I strongly believe this patch only affects Skylake hosts and “fixes” the three tests for them. I have no idea why Westmere emulation (?) is not successful on Skylake or why this assertion in qemu-kvm happens. Perhaps the two architectures are too different, not compatible. This is essentially just host CPU passthrough for Skylake, so I believe all users would do just that.

lenoch:
1. I second that, Lago itself (unlike the oVirt plugin) doesn't care about CPU family, only the model.
2. Yeah, just do host-passthrough unless overridden.

michalskrivanek:
Yeah, the only problem is that it's not any better and it covers real bugs.
If Westmere emulation CPU model in a nested VM on Skylake hw does not work then it is a QEMU/KVM bug and should treated as such. Lago is not the place to work around such issues - they need to be fixed for the sake of users using the nested virtualization feature.

mykaul:
On Feb 10, 2017 12:08 PM, "michalskrivanek" <notifications@github.com>
wrote:

Yeah, the only problem is that it's not any better and it covers real bugs.
If Westmere emulation CPU model in a nested VM on Skylake hw does not work
then it is a QEMU/KVM bug and should treated as such. Lago is not the place
to work around such issues - they need to be fixed for the sake of users
using the nested virtualization feature.


Was a bug on qemu-kvm opened?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/lago-project/lago/pull/448#issuecomment-278905296>, or mute
the thread

nvgoldin:
> I strongly believe this patch only affects Skylake hosts and “fixes” the three tests for them. 

@lenoch - The original problem reported by @mz-pdm was when we did use host-passthrough, though you might be right.

@mykaul - https://bugzilla.redhat.com/show_bug.cgi?id=1421430 (for Broadwell at least). 
@michalskrivanek, I agree, though considering the time it will probably take to fix such an issue, I think we do need to find a work around inside Lago, so again what I suggest is:
1. Default will be host-passthrough.
2. In OST, we'll add the explicit model we want(can be Westmere to begin with, similarly to now)
3. If nested virtualization fails for someone's specific setup - he can manually edit 'LagoInitFile' to find the combination which works for him.

I think this solution is generic enough to not introduce any work-around inside Lago, but would allow any user to configure a work-around if needed.

mz-pdm:
> If nested virtualization fails for someone's specific setup - he can manually edit 'LagoInitFile' to find the combination which works for him.

It would be nice if in such a case Lago detected the situation and provided the user a hint what to do (rather than just failing with some exception).

nvgoldin:
@mz-pdm, you know if there is any way to detect that? (without just blacklisting CPUs)?
I think ``virsh cpu-compare`` won't be useful for nested virtualization compare.

mz-pdm:
@nvgoldin, `virsh cpu-compare` works for me. E.g. if I run it in a L2 VM with Haswell-noTSX it tells me that it's incompatible with the host CPU, while Westmere is reported as OK.

nvgoldin:
upstream bug: https://bugs.launchpad.net/qemu/+bug/1665389
I think the bottom line there is that we need to switch back to 'host-passthrough', as nested virtualization is not supported unless using 'cpu -host' on the L0 host. (and again, allow overriding it if wanted, like in HE).

nvgoldin:
Hi @lenoch  - this should be fixed in https://github.com/lago-project/lago/pull/470(I added skylake too). Please tell me if it works for you. So closing this for now.

PR #449, opened 2017-02-08
testlib: handle non-callables as well as callables as test params

leongold:
As of now assert_equals* expects the first passed argument to be a
callable, however it is very likely we'd want to evaluate
non-callables as well, removing the need to continuously use
assert_true*(lambda: foo == True) constructs.

### DIFF
--- a/ovirtlago/testlib.py
+++ b/ovirtlago/testlib.py
@@ -21,7 +21,9 @@ import datetime
 import functools
 import logging
 import os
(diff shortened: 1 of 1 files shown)


leongold:
You're right. I envisioned obj.attr to be passed and re-evaluated, but it won't be re-evaluated unless we use some getattr(obj, 'attr') construct (which only makes things worse).

PR #452, opened 2017-02-15
Change lazy_refcounts to ON in the snapshots, move to virtio-console, add multiple queues to virtio-scsi

mykaul:
1. This in theory should improve performance. I did not measure much,
but I'm running in RAM / SSD, so can't expect much.
2. Hopefully virtio-console improves performance a bit over regular console.

### DIFF
--- a/lago/prefix.py
+++ b/lago/prefix.py
@@ -717,8 +717,10 @@ class Prefix(object):
             template_store.get_stored_metadata(template_version, ),
         )
         base = template_store.get_path(template_version)
-        qemu_cmd = ['qemu-img', 'create', '-f', 'qcow2', '-b', base, disk_path]
-
+        qemu_cmd = [
+            'qemu-img', 'create', '-f', 'qcow2', '-o', 'lazy_refcounts=on',
+            '-b', base, disk_path
+        ]
         return qemu_cmd, disk_metadata, base
 
     def _ova_to_spec(self, filename):
--- a/lago/dom_template.xml
+++ b/lago/dom_template.xml
@@ -17,6 +17,7 @@
       <acpi/>
       <apic/>
       <pae/>
+      <vmport state='off'/>
     </features>
     <devices>
       <emulator>@QEMU_KVM@</emulator>
@@ -36,11 +37,8 @@
         <backend model='random'>/dev/random</backend>
         <address type='pci' domain='0x0000' bus='0x00' slot='0x09' function='0x0'/>
       </rng>
-      <serial type='pty'>
-        <target port='0'/>
-      </serial>
       <console type='pty'>
-        <target type='serial' port='0'/>
+        <target type='virtio' port='0'/>
       </console>
       <video>
         <model type='cirrus' vram='16384' heads='1'/>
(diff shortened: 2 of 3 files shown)


mykaul:
@nvgoldin - looks like the Jenkins failure is unrelated to this patch:
07:20:52 No matching package to install: 'python2-paramiko >= 2.1.1'

mykaul:
ci test please

mykaul:
ci test please

nvgoldin:
@ifireball  - Could you take a look? I think there is a problem with the fc25 mirror:
```14:23:23 Failed to synchronize cache for repo 'fedora-updates-fc25', disabling.```
I checked manually the needed RPM is there.

**Comment by ifireball** [comment]
I fixed the mirror (with a workaround) for now.
I merged a patch today to get future syncs right

**Comment by nvgoldin** [comment]
@ifireball  - thanks.

**Comment by nvgoldin** [comment]
ci test please

File: lago/prefix.py [hunk 1]
```diff
@@ -645,7 +645,10 @@ def _handle_lago_template(
             template_store.get_stored_metadata(template_version, ),
         )
         base = template_store.get_path(template_version)
-        qemu_cmd = ['qemu-img', 'create', '-f', 'qcow2', '-b', base, disk_path]
+        qemu_cmd = [
+            'qemu-img', 'create', '-f', 'qcow2', '-o', 'compat=1.1', '-o',
+            'lazy_refcounts=on', '-b', base, disk_path
+        ]
```
nvgoldin: Do we also need to set the device cache definition to 'writethrough'? Couldn't figure if that is the default. I don't think we add any definition now(by dumping the xml from libvirt).
```
   "lazy_refcounts"
               If this option is set to "on", reference count updates are postponed with the goal of avoiding metadata I/O and improving performance. This is particularly interesting with
               cache=writethrough which doesn't batch metadata updates. The tradeoff is that after a host crash, the reference count tables must be rebuilt, i.e. on the next open an (automatic) "qemu-img
               check -r all" is required, which may take some time.
```
qemu-kvm:
```
By default, the cache=writeback mode is used. It will report data writes as completed as soon as the data is present in the host page cache.  This is safe as long as your guest OS makes sure to
           correctly flush disk caches where needed. If your guest OS does not handle volatile disk write caches correctly and your host crashes or loses power, then the guest may experience data
           corruption.
```

nvgoldin:
@mykaul - we'll have to wait a little before merging, until I fix https://ovirt-jira.atlassian.net/browse/OVIRT-113

lago/prefix.py
mykaul: I don't remember exactly why, but I could not change the cache mode. I'll try to look into it later.

mykaul:
@eedri - added removal of the IO threads to this commit. Now let's hope it's the only issue in -ev we've been using.

eedri:
+1, I still wonder how much benefit we'll gain from it, if the -ev is available on RHEV channels and CentOS VIRT repos, but let's see.

gbenhaim:
ci merge please

PR #456, opened 2017-02-26
[WIP] Adding hook support

gbenhaim:
1. The hooks are defined in the init file and
   can be any type of script (shell, python...).

2. The hooks will be run on the host which runs Lago.

3. Hooks will be only supported for functions that were
   decorated with "hooks.with_hooks" (for now just start / stop)


lago/hooks.py
```diff
@@ -0,0 +1,254 @@
+# coding=utf-8
+import functools
+import os
```
mykaul: Do we care about import order, alphabetically or something?

lago/hooks.py
```diff
@@ -0,0 +1,254 @@
...
+            return func(*args, **kwargs)
+
+        cmd = func.__name__
+        if cmd.startswith('do_'):
+            cmd = cmd[3:]
+
+        hooks = Hooks(prefix.paths.hooks())
+        hooks.run_pre_hooks(cmd)
+        result = func(*args, **kwargs)
+        hooks.run_post_hooks(cmd)
```
mykaul: The hooks run regardless of the result, and are not aware of the result. Perhaps worth passing it to them?

lago/hooks.py
```diff
@@ -0,0 +1,254 @@
...
+
+        if not hooks:
+            LOGGER.debug('No hooks were found for command: {}'.format(cmd))
+            return
+
+        # Avoid Recursion
+        try:
+            with utils.DirLockWithTimeout(cmd_dir):
+                self._run_hooks(
+                    sorted([path.join(hook_dir, hook) for hook in hooks])
```
mykaul: hooks are running alphabetically?

nvgoldin:
1. I think the hooks should be at the VM level. 
2. I think the hooks should be separated to different types: 
  * VM hooks - executed by Lago itself on the VM(by using SSH), similarly to 'deploy', and as result, possible only at post-start, pre-stop, etc.
  * Host hooks - executed on the host. For those I'm not 100% sure they are really needed, as this can just be wrapped in a bash command instead. It feels weird IMO that lago takes 'general' control of the host without direct relation to what it manages. Though I can understand the freedom it gives.
3. With 1 and 2, adding support for 'global' hooks, which are ran once per lago command, shouldn't be difficult.

nvgoldin:
Some possible use-cases of VM hooks:
1. Pre-stop: Generate SOS report before VM shut-down, dump postgresql db when running 'lago stop', etc.
2. Post-start: Install a service and assert it is running.

Some possible use-cases of Host hooks:
1. OST Pre-host-stop: Run API calls on the engine before stopping the VM.
2. OST Pre-host-stop: shut-down storage properly.

lago/hooks.py
gbenhaim: will fix.

lago/hooks.py
gbenhaim: 1. Post hook won't run if 'func' throws an exception, do we want to change this behavior ?
2. When passing 'result' to the post hook I need to assume that it is a string, or can be evaluated to a meaningful string, I'm not sure that I can take this assumption.

lago/hooks.py
gbenhaim: I want to preserve the order in which the hooks were mentioned in the init file, but now I realize that I didn't prefix them with an index. will fix.

PR #458, opened 2017-02-27
Adding basic video device

gbenhaim:
Some images will need a video device to boot,
so adding a simple video device.

### DIFF
--- a/lago/dom_template.xml
+++ b/lago/dom_template.xml
@@ -42,6 +42,11 @@
       <console type='pty'>
         <target type='serial' port='0'/>
       </console>
+      <video>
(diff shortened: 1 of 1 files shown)


gbenhaim:
ci merge please

lago/dom_template.xml
```diff
@@ -42,6 +42,11 @@
       <console type='pty'>
         <target type='serial' port='0'/>
       </console>
+      <video>
+        <model type='cirrus' vram='16384' heads='1'/>
```
mykaul: Cirrus is a bad choice - we've just switched away from it in oVirt.
Should be VGA.

Issue #307, opened 2016-09-06
lago destroy doesn't work when init filed


kivio:
I have problem with finish lago init job.  (it hangs on downloading template without any timeout exception)
If i break this process and want to clear environment lago destroy raise exception. 

```
[kivio@centos lago-test]$ lago destroy
current session does not belong to lago group.
Error occured, aborting
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/lago/cmd.py", line 691, in main
    cli_plugins[args.verb].do_run(args)
  File "/usr/lib/python2.7/site-packages/lago/plugins/cli.py", line 180, in do_run
    self._do_run(**vars(args))
  File "/usr/lib/python2.7/site-packages/lago/utils.py", line 471, in wrapper
    workdir_path = workdir_class.resolve_workdir_path(workdir_path)
  File "/usr/lib/python2.7/site-packages/lago/workdir.py", line 375, in resolve_workdir_path
    os.path.abspath(start_path)
RuntimeError: Unable to find workdir for /home/kivio/lago-test
```

Only one way to retry it is manually removing .lago directory.

nvgoldin:
Hi @kivio , can you please provide a way to replicate the issue?

nvgoldin:
OK - re-read what you wrote and I understand what you mean. I'm changing the title, as this will happen with any verb, not only 'destory', the problem is that when ``lago init`` fails it pollutes the enviornment and currently the only solution is to delete manually the ``.lago`` directory. A solution would be to delete it when ``init`` fails. Fixing it require extra-attention to the corner-cases(for example running ``init`` twice in a row)

PR #466, opened 2017-03-03
Don't leave malformed workdir on 'lago init' failure

nvgoldin:
1. Fail early on 'init' command if no configuration file exists.
2. Delete the workdir if it is empty, and 'init' failed to create the
prefix.

This resolves an annoying bug that when 'init' failed, you had to remove
the '.lago' directory manually.

closes: https://github.com/lago-project/lago/issues/307


nvgoldin:
Thanks for the review. 
SO - I'm merging this, if anyone could help testing it would be good -as this involves some `os.unlink` commands. I tested it on several scenarios with few prefixes to ensure it doesn't detect none-malformed workdirs as malformed, but still - better be safe, than really sorry ;-)

nvgoldin:
ci merge please

PR #474, opened 2017-03-13
Allow to call sysprep.update() which will update all packages

mykaul:
during sysprep.
For the time being, no one is calling it - I'm using it manually
by modifying vm.py if I need to, but it's good to have it.


gbenhaim:
ci merge please

Issue #473, opened 2017-03-13
Auto generate the init file of exported env


gbenhaim:
When exporting vm disk's from an env, export an init file as well that can be used later to recreate the env.

PR #346, opened 2016-10-23
Ensure consistent load order of domains and networks.

mykaul:
By processing JSON or YAML files via Ordered dicts.,
we ensure that domains are processed in the order they are specified
in the relevant configuration files.

PR #477, opened 2017-03-14
Generate an init file that can be used with exported images

gbenhaim:
The generated init file will represent the env and will contain
relative paths to the exported images.

closes https://github.com/lago-project/lago/issues/473

### DIFF
--- a/lago/plugins/vm.py
+++ b/lago/plugins/vm.py
@@ -27,6 +27,7 @@ The other plugin extension point, the [VM Provider Plugin], that allows you to
 create an alternative implementation of the provisioning details for the VM,
 for example, using a remote libvirt instance or similar.
 """
+from copy import deepcopy
 import contextlib
 import functools
 import logging
@@ -363,6 +364,10 @@ class VMPlugin(plugins.Plugin):
     @property
     def disks(self):
         return self._spec['disks'][:]
+
+    @property
+    def spec(self):
+        return deepcopy(self._spec)
 
     def name(self):
         return str(self._spec['name'])
--- a/lago/prefix.py
+++ b/lago/prefix.py
@@ -961,7 +961,8 @@ class Prefix(object):
                     'dev': disk['dev'],
                     'format': disk['format'],
                     'metadata': metadata,
-                    'type': disk['type']
+                    'type': disk['type'],
+                    'name': disk['name']
                 },
             )
 
@@ -1009,9 +1010,14 @@ class Prefix(object):
             self.save()
             rollback.clear()
 
-    def export_vms(self, vms_names, standalone, export_dir, compress):
-
-        self.virt_env.export_vms(vms_names, standalone, export_dir, compress)
+    def export_vms(
+        self, vms_names, standalone, export_dir, compress, init_file_name,
+        out_format
+    ):
+        self.virt_env.export_vms(
+            vms_names, standalone, export_dir, compress, init_file_name,
+            out_format
+        )
 
     def start(self, vm_names=None):
         """
--- a/lago/utils.py
+++ b/lago/utils.py
@@ -690,6 +690,83 @@ def get_hash(file_path, checksum='sha1'):
     return sha.hexdigest()
 
 
+def filter_spec(spec, paths, wildcard='*', separator='/'):
+    """
+    Remove keys from a spec file.
+    For example, with the following path: domains/*/disks/*/metadata
+    all the metadata dicts from all domains disks will be removed.
+
+    Args:
+        spec (dict): spec to remove keys from
+        paths (list): list of paths to the keys that should be removed
+        wildcard (str): wildcard character
+        separator (str): path separator
+
+    Returns:
+        None
+
+    Raises:
+        utils.LagoUserException: If a malformed path was detected
+    """
+
+    def remove_key(path, spec):
+        if len(path) == 0:
+            return
+        elif len(path) == 1:
+            key = path.pop()
+            if not isinstance(spec, collections.Mapping):
+                raise LagoUserException(
+                    'You have tried to remove the following key - "{key}".\n'
+                    'Keys can not be removed from type {spec_type}\n'
+                    'Please verify that path - "{{path}}" is valid'.format(
+                        key=key, spec_type=type(spec)
+                    )
+                )
+            if key == wildcard:
+                spec.clear()
+            else:
+                spec.pop(key, None)
+        else:
+            current = path[0]
+            if current == wildcard:
+                if isinstance(spec, list):
+                    iterator = iter(spec)
+                elif isinstance(spec, collections.Mapping):
+                    iterator = spec.itervalues()
+                else:
+                    raise LagoUserException(
+                        'Glob char {char} should refer only to dict or list, '
+                        'not to {spec_type}\n'
+                        'Please fix path - "{{path}}"'.format(
+                            char=wildcard, spec_type=type(spec)
+                        )
+                    )
+
+                for i in iterator:
+                    remove_key(path[1:], i)
+            else:
+                try:
+                    remove_key(path[1:], spec[current])
+                except KeyError:
+                    raise LagoUserException(
+                        'Malformed path "{{path}}", key "{key}" '
+                        'does not exist'.format(key=current)
+                    )
+                except TypeError:
+                    raise LagoUserException(
+                        'Malformed path "{{path}}", can not get '
+                        'by key from type {spec_type}'.
+                        format(spec_type=type(spec))
+                    )
+
+    for path in paths:
+        try:
+            remove_key(path.split(separator), spec)
+        except LagoUserException as e:
+            e.message = e.message.format(path=path)
+            raise
+
+
 class LagoException(Exception):
     pass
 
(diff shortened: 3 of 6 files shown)


lago/plugins/vm.py
```diff
@@ -27,6 +27,7 @@
 create an alternative implementation of the provisioning details for the VM,
 for example, using a remote libvirt instance or similar.
 """
+from copy import deepcopy
```
nvgoldin: We have our "specific" implementation of deepcopy in 'utils.py', I guess here its safe? (it was added to avoid the cross-references issue iirc)

lago/plugins/vm.py
```diff
@@ -364,6 +365,10 @@ def metadata(self):
     def disks(self):
         return self._spec['disks'][:]
 
+    @property
+    def spec(self):
```
nvgoldin: This makes the new spec somewhat "read only", no? was that the intention?

lago/utils.py
```diff
@@ -690,6 +690,65 @@ def get_hash(file_path, checksum='sha1'):
...
 
 
+def filter_spec(spec, paths):
+    """
+    Remove keys from a spec file.
+    For example, with the following path: domains/*/disks/*/metadata
+    all the metadata dicts from all domains disks will be removed.
+
+    Args:
+        spec (dict): spec file to remove keys from
```
nvgoldin: Its a dict(not a file) :) as you wrote..

lago/utils.py
```diff
@@ -690,6 +690,65 @@ def get_hash(file_path, checksum='sha1'):
...
+        paths (list): list of paths to the keys that should be removed
+
+    Returns:
+        None
```
nvgoldin: Please add 'raises' here(look elsewhere in the file the exact syntax so it would format sphinixily nicely)

lago/utils.py
```diff
@@ -690,6 +690,65 @@ def get_hash(file_path, checksum='sha1'):
...
+    """
+
+    def remove_key(path, spec):
+        if len(path) == 0:
+            return
+        elif len(path) == 1:
+            key = path.pop()
+            if not isinstance(spec, dict):
+                raise LagoUserException(
```
nvgoldin: I would print a more detailed message here, that hints what to do or what the user did wrong.

lago/utils.py
```diff
@@ -690,6 +690,65 @@ def get_hash(file_path, checksum='sha1'):
...
+            current = path[0]
+            if current == '*':
+                if isinstance(spec, list):
+                    iterator = iter(spec)
+                elif isinstance(spec, dict):
+                    iterator = spec.itervalues()
+                else:
+                    raise LagoUserException(
+                        'Glob char * should refer only to dict or list, '
+                        'not to {}'.format(type(spec))
```
nvgoldin: Maybe add indication in the message where it broke.

lago/utils.py
```diff
@@ -690,6 +690,65 @@ def get_hash(file_path, checksum='sha1'):
...
+            else:
+                try:
+                    remove_key(path[1:], spec[current])
+                except KeyError:
+                    raise LagoUserException(
+                        'Malformed path "{}", key "{}" '
+                        'does not exist'.format('/'.join(path), current)
+                    )
+                except TypeError:
+                    raise LagoUserException(
```
nvgoldin: when is this condition met and why? and how is it different from line 726?

lago/virt.py
```diff
@@ -219,6 +221,82 @@ def export_vms(self, vms_names, standalone, dst_dir, compress):
...
+                    'domains/*/disks/*/metadata',
+                    'domains/*/metadata/deploy-scripts', 'domains/*/snapshots',
+                    'domains/*/name', 'nets/*/mapping'
+                ]
+            spec = self.get_env_spec(filters)
+
+            for _, domain in spec['domains'].viewitems():
+                for disk in domain['disks']:
+                    if disk['type'] == 'template':
+                        disk['template_type'] = 'qcow2'
```
nvgoldin: Not sure if this is the right place for this logic. Should probably be a different method.

lago/virt.py
```diff
@@ -219,6 +221,82 @@ def export_vms(self, vms_names, standalone, dst_dir, compress):
...
+                    elif disk['type'] == 'empty':
+                        disk['type'] = 'file'
+                        disk['make_a_copy'] = 'True'
+
+                    # Insert the relative path to the exported images
+                    disk['path'] = os.path.join(
+                        '$LAGO_INITFILE_PATH', os.path.basename(disk['path'])
+                    )
+
+            with open(os.path.join(dst_dir, 'LagoInitFile'), 'wt') as f:
```
nvgoldin: The dumping filename should be configuable, moreover it should drop it to stdout by default(like ansible_hosts cmd), so you could do something like:
```
lago dump-init >> my-init-el73-2011
```

Not sure if its an issue but if the file already exists(LagoInitFile) it will override it.

lago/prefix.py
```diff
@@ -961,7 +961,8 @@ def _create_disks(
                     'dev': disk['dev'],
                     'format': disk['format'],
                     'metadata': metadata,
-                    'type': disk['type']
+                    'type': disk['type'],
+                    'name': disk['name']
```
nvgoldin: :+1:

lago/utils.py
```diff
@@ -690,6 +690,65 @@ def get_hash(file_path, checksum='sha1'):
...
+            key = path.pop()
+            if not isinstance(spec, dict):
+                raise LagoUserException(
+                    'Can not remove key "{}" from "{}" {}'.
+                    format(key, spec, type(spec))
+                )
+            spec.pop(key, None)
+        else:
+            current = path[0]
+            if current == '*':
```
nvgoldin: probably better to add this as a default to the function:
```
def .... (..., sep='*')
```

lago/plugins/vm.py
gbenhaim: What do you mean by "cross-references" ?

lago/plugins/vm.py
gbenhaim: Yes.
Any change to the spec will effect the structure of the env. IMO changes to the env should be done with specific object's methods (in this case the 'vm' object), and not directly to its data.

lago/plugins/vm.py
nvgoldin: See the discussion here: https://github.com/lago-project/lago/pull/346#pullrequestreview-5799308

Not 100% if its relevant, just checking.

lago/plugins/vm.py
nvgoldin: 👍

lago/virt.py
gbenhaim: This method is for generating an init file that can be used with the exported images 'out of the box'.
For general spec you should use ```get_env_spec```

lago/virt.py
gbenhaim: This method is called only when exporting images, so it makes sense to write directly to a file.
I agree that the filename should be configurable.

lago/utils.py
gbenhaim: Example for each exception (by the order I catch them):

```
You have tried to remove the following key - "template".
Keys can not be removed from type <type 'list'>
Please verify that path - "domains/*/disks/template" is valid
```

```
Glob char * should refer only to dict or list, not to <type 'int'>
Please fix path - "domains/*/memory/*/some_key"
```

```
Malformed path "domains/*/gpu/*", key "gpu" does not exist
```

```
Malformed path "domains/*/disks/root/vda", can not get by key from type <type 'list'>
```

lago/utils.py
gbenhaim: done

lago/utils.py
gbenhaim: done

lago/utils.py
gbenhaim: done

lago/utils.py
gbenhaim: done

lago/utils.py
gbenhaim: done

lago/utils.py
```diff
@@ -690,42 +690,56 @@ def get_hash(file_path, checksum='sha1'):
...
+        utils.LagoUserException: If a malformed path was detected
     """
 
     def remove_key(path, spec):
         if len(path) == 0:
             return
         elif len(path) == 1:
             key = path.pop()
-            if not isinstance(spec, dict):
+            if not isinstance(spec, collections.Mapping):
```
nvgoldin: Cool. This should work for OrderedDicts also?

lago/utils.py
```diff
@@ -690,42 +690,56 @@ def get_hash(file_path, checksum='sha1'):
...
                 raise LagoUserException(
-                    'Can not remove key "{}" from "{}" {}'.
-                    format(key, spec, type(spec))
+                    'You have tried to remove the following key - "{key}".\n'
+                    'Keys can not be removed from type {spec_type}\n'
+                    'Please verify that path - "{{path}}" is valid'.format(
```
nvgoldin: Nice.

gbenhaim:
ci merge please

gbenhaim:
ci test please

gbenhaim:
ci merge please

Issue #464, opened 2017-03-01
Add 'lago shutdown'


gbenhaim:
This command should shutdown (poweroff) the vms (not just stop them).
Could be implemented using libvirt api or with sending poweroff command through ssh.
We need to decide if to stop the networks as well, or to leave this part to ```lago stop```.
It will nice to add ```-r``` option for restart.

PR #480, opened 2017-03-16
Adding lago shutdown

gbenhaim:
1. Adding lago shutdown - gracefully shutdown / reboot vm.
2. If acpi is enabled in the vm, use libvirt function, otherwise
   try to run the command using ssh.
3. Change VirtEnv.get_vms to support filtering by name.

closes https://github.com/lago-project/lago/issues/464

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -278,6 +278,30 @@ def do_stop(prefix, vm_names, **kwargs):
 
 
 @lago.plugins.cli.cli_plugin(
+    help='Shutdown and destroy, or reboot vms',
+    description='This command will shutdown or reboot the given vms. '
+    'Networks that will not have running vms connected to them after '
+    'running this command will be stopped as well.'
+)
+@lago.plugins.cli.cli_plugin_add_argument(
+    '-r',
+    '--reboot',
+    help='If specified, reboot the requested vms',
+    action='store_true',
+)
+@lago.plugins.cli.cli_plugin_add_argument(
+    'vm_names',
+    help='Name of the vms to shutdown',
+    metavar='VM_NAME',
+    nargs='*',
+)
+@in_lago_prefix
+@with_logging
+def do_shutdown(prefix, vm_names, reboot, **kwargs):
+    prefix.shutdown(vm_names, reboot)
+
+
+@lago.plugins.cli.cli_plugin(
     help='Export virtual machine disks',
     description='This command will export the disks of the given vms. '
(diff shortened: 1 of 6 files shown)


lago/cmd.py
```diff
@@ -277,6 +277,25 @@ def do_stop(prefix, vm_names, **kwargs):
     prefix.stop(vm_names=vm_names)
 
 
+@lago.plugins.cli.cli_plugin(help='Shutdown and destroy lago resources')
+@lago.plugins.cli.cli_plugin_add_argument(
+    '-r',
+    '--reboot',
```
mykaul: reboot is not shutdown?

lago/virt.py
```diff
@@ -332,14 +332,29 @@ def start(self, vm_names=None):
                     rollback.prependDefer(vm.stop)
                 rollback.clear()
 
-    def stop(self, vm_names=None):
+    def _get_stoppable_vms_and_nets(self, vm_names):
```
mykaul: I'm not sure if I wouldn't have done get_stoppable_vms() and get_stoppable_nets()

Also, the word 'stoppable' is a bit strange. _get_running_nets() and _get_running_vms() perhaps?

lago/virt.py
```diff
@@ -376,6 +415,36 @@ def get_net(self, name=None):
...
+        for name in vm_names:
+            try:
+                vms[name] = self._vms[name]
+            except KeyError:
+                # TODO: add resolver by suffix
+                missing_vms.append(name)
+
+        if missing_vms:
+            raise utils.LagoUserException(
+                'The following vms does not exist: \n{}'.
```
mykaul: does not -> do not

lago/cmd.py
gbenhaim: No.
Shutdown will keep the vm off, while reboot will turn it off and then power it on.
This is equivalent to the shell command ```shutdown```.

mykaul: So the plugin help says 'Shutdown and destroy' - how does that work with reboot?

gbenhaim: Changed the help message.

lago/virt.py
gbenhaim: I've split this function and rename it. I hope it's clear now.

lago/virt.py
gbenhaim: Done.

gbenhaim:
ci merge please

PR #490, opened 2017-03-20
Let us control if DNS should be enabled in a libvirt network.

lenoch:
This is a simple way to make sure we can use DHCP in multiple networks (with only one also served by DNS).

### DIFF
--- a/lago/net_nat_template.xml
+++ b/lago/net_nat_template.xml
@@ -6,7 +6,7 @@
     </nat>
   </forward>
   <bridge name='@BR_NAME@' stp='on' delay='0' />
-  <dns forwardPlainNames='yes' >
+  <dns enable='@ENABLE_DNS@' forwardPlainNames='yes'>
   </dns>
   <ip address='@GW_ADDR@' netmask='255.255.255.0' />
   <ip family='ipv6' address='fd8f:1391:3a82:@SUBNET@::1' prefix='64' />
--- a/lago/virt.py
+++ b/lago/virt.py
@@ -650,10 +650,15 @@ class NATNetwork(Network):
 
         subnet = self.gw().split('.')[2]
         replacements = {
-            '@NAME@': self._libvirt_name(),
+            '@NAME@':
+                self._libvirt_name(),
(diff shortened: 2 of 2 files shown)


lenoch:
ci test please

mykaul:
I don't understand the issue. DHCP was not working on multiple networks? DNS I believe had an issue with multiple host names - how the interfaces are named, but that's about it.

lenoch:
@mykaul This lets allows me to disable DNS in all networks but the management network, see https://gerrit.ovirt.org/#/c/74360/ It is necessary that only the management network supplies DNS configuration to hosts (via DHCP leases) because we would lose any information about nameservers otherwise (and this is why migrations fail, as destination hosts are chosen by name, not just IP, so we definitely need a working resolver).

More detailed explanation:
In OST, 003_basic_networking also enables DHCP on the VM_Network. Now that I am removing VLAN tagging from the network (thanks to moving it to a separate NIC) it finally works. But once a DHCP lease is received for this NIC, DNS configuration changes on the host. Ultimately, it is wiped off. One way to overcome this is to allow DNS only in one network.

lenoch:
In any case, I would like to have an option to turn off DNS, even if we end up solving the above problem in a different way. I really believe this patch is harmful, as always :-D

gbenhaim:
It won't make any harm to have this option.
ci merge please

gbenhaim:
ci merge please

mykaul:
@lenoch , 
For next time: 
1. if 'dns_domain_name' is defined, we probably want DNS.
2. Don't change the default. The default was to have DNS, now you've changed it to NOT having DNS.

lenoch:
@mykaul The libvirt default (`yes`) is preserved (currently, `enable_dns` is not used anywhere, and thus we get `True` from the `.get()`).

It seems like a good idea to me to let `dns_domain_name` into the play, I only didn't want to chase off my reviewers by introducing logic more complex than necessary, given that this was a last-minute addition.

```
'@ENABLE_DNS@':
+                'yes' if self._spec.get('enable_dns', True) else 'no',
```

PR #493, opened 2017-03-21
Adding DHCP reservation to unmanaged networks

gbenhaim:
At this moment DHCP reservation is available only for networks that
were marked as management in the init file.

This patch add the option to specify reserved addresses in unmanged
networks as well.

### DIFF
--- a/lago/virt.py
+++ b/lago/virt.py
@@ -701,24 +701,24 @@ class NATNetwork(Network):
                 )
             )
 
-            if self.is_management():
-                for hostname, ip4 in self._spec['mapping'].items():
-                    dhcp.append(
-                        lxml.etree.Element(
-                            'host',
-                            mac=utils.ipv4_to_mac(ip4),
-                            ip=ip4,
-                            name=hostname
-                        )
+            for hostname, ip4 in self._spec['mapping'].items():
+                dhcp.append(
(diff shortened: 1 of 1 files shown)


gbenhaim:
@mykaul @nvgoldin @ifireball thoughts ?

mykaul:
What will be the hostname for those non-management networks? I know it's something different (see prefix.py - '%s-eth%d' % (dom_name, idx) - and it's not reflected in this patch. I'm OK with hostname-<networkname> as well, btw.

gbenhaim:
@mykaul Do we want that the hostname will be the same across all networks ? (personally I don't see a reason why not)

mykaul:
@gbenhaim - we cannot have the same hostname across networks, otherwise it may resolve a hostname to an IP that is not on the network you expect it to be. For example, engine FQDN will not be resolved to the IP on the mgmt network but on the storage network, etc.

nvgoldin:
@lenoch - Can you help reviewing this change? Is there anything we might be missing(except the DNS)?
The functionality we require is that none-managment nets will have "fixed" DHCP leases, we want them to have the same IP over consecutive runs(lago start && lago stop && lago start).

gbenhaim:
Implemented in https://github.com/lago-project/lago/pull/500

PR #499, opened 2017-04-03
tests: add prefix and cpu tests to ovirt.runtest

nvgoldin:
1. Run a few "simple" prefix API calls, as used by OST.

2. Test 'cpu_model' definition. This is not strictly related to ovirtlago
but we better do this on a standard distribution(rather than
CirrOS), so in order not to make the tests longer, it is here.


nvgoldin:
ci merge please

nvgoldin:
ci merge please

PR #500, opened 2017-04-04
Move to a single DNS server per management network 

nvgoldin:
1. Select a network to serve as a central DNS server, by the following logic:
  - if it is the only network defined in the init file, else;
  - the only management network defined in the init file, else;
  - if there is more than one management network, use the one with DNS
    configuration in the init file; else;
  - if there is no DNS configuration, choose the first management network found.

Corner cases such as defining DNS for two different 'management'
networks are not allowed and will throw an error on 'lago init'.

2. Once the network is chosen, all DNS records will be configured in it,
all other networks will forward the requests to it.

3. Custom DNS records can also be defined in the init file, by adding
'dns_records' in the network that has DNS configuration.

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -936,7 +936,7 @@ def main():
 
     try:
         cli_plugins[args.verb].do_run(args)
-    except utils.LagoUserException as e:
+    except utils.LagoException as e:
         LOGGER.info(e.message)
         LOGGER.debug(e)
         sys.exit(2)
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -456,7 +456,9 @@ def do_shell(prefix, host, args=None, **kwargs):
     sys.exit(result.code)
 
 
-@lago.plugins.cli.cli_plugin(help='Open serial console to the domain', )
+@lago.plugins.cli.cli_plugin(
+    help='Open serial console to the domain',
+)
 @lago.plugins.cli.cli_plugin_add_argument(
     'host',
     help='Host to connect to',
--- a/lago/export.py
+++ b/lago/export.py
@@ -152,9 +152,8 @@ class TemplateExportManager(DiskExportManager):
     """
 
     def __init__(self, dst, disk_type, disk, do_compress, *args, **kwargs):
-        super(TemplateExportManager, self).__init__(
-            dst, disk_type, disk, do_compress
-        )
+        super(TemplateExportManager,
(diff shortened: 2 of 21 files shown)


ifireball:
The rules above don't provide proper handling to the situation where we have two management networks each connected to a different set of VMs.

nvgoldin:
Thanks for the input, after off-line discussion I plan to do the following changes:
1. Ensure each domain has one and only one management network configured, otherwise fail the initialization stage.
2. Each management network will have *all* DNS  enteries added(so if there are 2 management networks, they will both have the DNS records of each other, despite unrelated), in order to avoid complexity of finding the connected components for each management network. I don't believe it adds any major overhead.
3. All *none* management network will be configured with:
  - in libvirt >= 2.2.0, dns disabled(not available in RHEL right now).
  - in libvirt < 2.2.0, will forward the requests to one of the management networks. This is not ideal, but will work for now, because we use NAT. Note that I did not find any other, official or none-official, way to disable DNS.

Few other corner cases:
1. If no 'management'  network is defined in the init file - it will select the first one.
2. If a VM will be configured without any management networks, it will fail.
3. If only one network is configured in the init file, it will be selected as management(even if not written explicitly).

nvgoldin:
PR Version 2
-----------------
 Move to a single DNS server per management network

1. Each management network will serve as a DNS server. None management
networks are configured in the following manner:
  - For libvirt version < 2.2, they will forward the DNS requests to one
    of the management networks.
  - For libvirt version >= 2.2, DNS is disabled.

2. Each VM can have one and only one management network, if it has more
than one, then 'init' stage will fail.

3. DNS records format:
  - 'domain_name-eth#' - where '#' is the index of the nic as declared in
    the  init file.
  - Additional record of 'domain_only' is mapped to the IP served by the
    management network.

4. If only one network is defined, it will be assumed as the management
one.

5. Custom DNS records can also be appended in the init file, by adding
'dns_records: {host:ip}' in any management network.


Also note that this PR forced me solve the the 'finding the management of the VM issue' - we can now use:
```python
host.get_mgmt()
```
Also, the management network no longer has to be defined first in the VM's nic lists.

nvgoldin:
ci test please

ovirt-infra:
Build finished.

nvgoldin:
ci test please

nvgoldin:
ci merge please

PR #506, opened 2017-04-09
Stop using sys.exc_info in LogTask

gbenhaim:
1. sys.exc_info returns the last exception that was caught,
this can confuse LogTask and make it print messages on exceptions that
were already handled and don't interupt to the flow of lago,
(for example when waiting for ssh).
Insted we will use the default arguments which are being passed
to __exit__, and reflects the last exception that was thrown and not handled.

2. If an error occurred, print that stack trace in LogTask.

### DIFF
--- a/lago/log_utils.py
+++ b/lago/log_utils.py
@@ -24,7 +24,7 @@ import logging
 import logging.config
 import os
 import re
-import sys
+import traceback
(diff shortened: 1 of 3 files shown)


gbenhaim:
ci merge please

Issue #508, opened 2017-04-10
ovirtlago: lago ovirt start sometimes fail with Fault reason is "Operation Failed". 


nvgoldin:
```
> lago --loglevel=debug ovirt start
Looking for a workdir
Checking if /home/ngoldin/src/nvgoldin.github.com/lago-init-files/lago_demo_imgs is a workdir
@ Starting oVirt environment: 
  # Start Prefix: 
    * Start nets: 
    * Start nets: Success (in 0:00:00)
    * Start vms: 
    * Start vms: Success (in 0:00:00)
  # Start Prefix: Success (in 0:00:00)
  # Activating Engine Hosts: 
  # Activating Engine Hosts: ERROR (in 0:00:01)
@ Starting oVirt environment: ERROR (in 0:00:01)
Error occured, aborting
Traceback (most recent call last):
  File "/home/ngoldin/src/nvgoldin.github.com/lago/ovirtlago/cmd.py", line 319, in do_run
    self.cli_plugins[args.ovirtverb].do_run(args)
  File "/home/ngoldin/src/nvgoldin.github.com/lago/lago/plugins/cli.py", line 184, in do_run
    self._do_run(**vars(args))
  File "/home/ngoldin/src/nvgoldin.github.com/lago/lago/utils.py", line 495, in wrapper
    return func(*args, **kwargs)
  File "/home/ngoldin/src/nvgoldin.github.com/lago/lago/utils.py", line 506, in wrapper
    return func(*args, prefix=prefix, **kwargs)
  File "/home/ngoldin/src/nvgoldin.github.com/lago/ovirtlago/cmd.py", line 232, in do_ovirt_start
    prefix.virt_env.engine_vm().start_all_hosts()
  File "/home/ngoldin/src/nvgoldin.github.com/lago/ovirtlago/utils.py", line 145, in wrapped_func
    return func(*args, **kwargs)
  File "/home/ngoldin/src/nvgoldin.github.com/lago/ovirtlago/virt.py", line 370, in start_all_hosts
    host_service.activate()
  File "/home/ngoldin/virtualenv/lago-venv/lib/python2.7/site-packages/ovirtsdk4/services.py", line 36984, in activate
    self._check_action(response)
  File "/home/ngoldin/virtualenv/lago-venv/lib/python2.7/site-packages/ovirtsdk4/service.py", line 129, in _check_action
    Service._raise_error(response, result.fault)
[...]

PR #511, opened 2017-04-19
ovirtlago: assert ovirt-engine and vdsmd services are running

nvgoldin:
In 'lago ovirt start', before making any API calls, assert that
ovirt-engine and vdsmd services report running on all hosts. Note that
the time includes the time we have to wait for the VMs to be up(and
sshd) available, and the time for the services to come up.

fixes: https://github.com/lago-project/lago/issues/508
and probably other sporadic issues.

### DIFF
--- a/ovirtlago/cmd.py
+++ b/ovirtlago/cmd.py
@@ -228,6 +228,10 @@ def do_ovirt_status(prefix, **kwargs):
 def do_ovirt_start(prefix, with_vms, **kwargs):
     with LogTask('Starting oVirt environment'):
         prefix.start()
+        with LogTask('Waiting for ovirt-engine status'):
+            prefix.virt_env.assert_engine_alive()
+        with LogTask('Waiting for vdsmd status'):
+            prefix.virt_env.assert_vdsm_alive()
         with LogTask('Activating Engine Hosts'):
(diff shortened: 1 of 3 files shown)


nvgoldin:
Verified locally with OST images.
@gbenhaim, @mykaul  - any concerns with this? Sort of need to get it in today to make it for v0.37

mykaul:
A bit sad it's needed (it's duplicating quite some code from ovirt-system-tests), but nothing wrong with the code, I guess.

Ideally, you should check engine health by looking at the Health Status page, btw, not just the service is up.

nvgoldin:
@mykaul  - thanks for the quick reply. 
Well I actually added the 'ovirt-engine' check while at it, the true problem was vdsmd. It is just not right to assume that 'vdsmd' is running when 'lago start' finishes. From what I tested it can take up to 40 seconds for that to happen(VM boot + vdsmd is running). Some API calls failed because the host VM was not even reachable in SSH nor console at the time we tried to run to activate the hosts.

As we "rise" from an unnatural state where engine + hosts are all shutdown, I think checking the services are up is preliminary to checking the Health status.

nvgoldin:
ci merge please

PR #513, opened 2017-04-22
 Make DNS changes introduced in 6386bc8 backward compatible

nvgoldin:
Commit 6386bc8 relays on DNS records stored in the Prefix metadata.
This commit ensures that environments which were initialized prior
to that commit, will work as before.

### DIFF
--- a/lago/providers/libvirt/network.py
+++ b/lago/providers/libvirt/network.py
@@ -34,7 +34,7 @@ LogTask = functools.partial(log_utils.LogTask, logger=LOGGER)
 
 
 class Network(object):
-    def __init__(self, env, spec):
+    def __init__(self, env, spec, compat):
         self._env = env
         libvirt_url = config.get('libvirt_url')
         self.libvirt_con = libvirt_utils.get_libvirt_connection(
@@ -42,6 +42,7 @@ class Network(object):
             libvirt_url=libvirt_url,
         )
         self._spec = spec
+        self.compat = compat
 
     def name(self):
         return self._spec['name']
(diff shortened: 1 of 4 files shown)


lago/virt.py
```diff
@@ -507,3 +507,10 @@ def get_snapshots(self, domains=None):
             snapshots[vm_name] = vm._spec['snapshots']
 
         return snapshots
+
+    def get_compat(self):
```
gbenhaim: Good idea!

lago/providers/libvirt/network.py
```diff
@@ -227,24 +228,48 @@ def make_ipv4(last):
...
+            if self.is_management():
+                domain_xml = ET.Element(
+                    'domain',
+                    name=self._spec['dns_domain_name'],
+                    localOnly='yes'
                 )
-            )
+                net_xml.append(domain_xml)
+                net_xml.append(
+                    self._generate_main_dns(self._spec['dns_records'], subnet)
```
gbenhaim: We don't want to forward plain names ?

nvgoldin: I think what makes sense is:
For all management networks - don't forward unqualified names, as we know this is the last endpoint and we probably don't want any unqualified DNS records outside the virtual network.
For all none-mgmt networks - forward(as they don't have DNS).

Prior to 6386bc8 we had forwardPlainNames='yes' in all networks(hardcoded in the template).

gbenhaim: +1

nvgoldin:
ci merge please

Issue #504, opened 2017-04-09
el6 machines fail to boot


nvgoldin:
It was reported by @danielBelenky  that el6 machines fail to boot, I can replicate it on my env with this simple init file:
```yaml
domains:
  lago-basic-suite-3-6-engine:
    vm-type: ovirt-engine
    memory: 4096
    nics:
      - net: lago-basic-suite-3-6-net-management
    disks:
      - template_name: el6-base
        type: template
        name: root
        dev: vda
        format: qcow2
    metadata:
      ovirt-engine-password: 123


nets:
  lago-basic-suite-3-6-net-management:
    type: nat
    dhcp:
      start: 100
      end: 254
    management: true
    dns_domain_name: lago.local
```

Investigating.

nvgoldin:
1. Looks like https://github.com/lago-project/lago/commit/179f5114fdca004fd8a0b8e40e6a166813a5a86d is the cause, for some reason el6 doesn't behave nicely with the virtio-console.
2. It was reported that after a while(around 15 minutes) the VM does actually come up, unfortunately I was unable to reproduce that locally yet in order to extract the logs.

mykaul:
Perhaps EL6 does not have virtio-console devices?

(Why do we care about EL6 anyway?)

nvgoldin:
> Perhaps EL6 does not have virtio-console devices?

Checking.. Anyways, Instead of reverting, I'll try diverging the el6 domain XML to a different template, so we don't have to use EL6 compatibility as a common-ground.

> (Why do we care about EL6 anyway?)

OST basic-suite-3.6 uses EL6 for the engine(host are already EL7).

mykaul:
We can remove OST basic-suite-3.6.

PR #515, opened 2017-04-23
Use a seperate template for el6

nvgoldin:
The idea is to "freeze" the el6 template, so we won't be breaking stuff when adding new features to the "upstream" template. Specifically for now, set the console back to 'serial'.

Fixes: https://github.com/lago-project/lago/issues/504

### DIFF
--- a/lago/providers/libvirt/templates/dom_template-el6.xml
+++ b/lago/providers/libvirt/templates/dom_template-el6.xml
@@ -1,0 +1,46 @@
+<domain type='kvm'>
+    <name>@NAME@</name>
+    <memory unit='MiB'>@MEM_SIZE@</memory>
+    <iothreads>1</iothreads>
+    <os>
+      <type arch='x86_64' machine='pc'>hvm</type>
+      <bootmenu enable='no'/>
+    </os>
+    <features>
+      <acpi/>
+      <apic/>
+      <pae/>
+      <vmport state='off'/>
+    </features>
+    <devices>
+      <emulator>@QEMU_KVM@</emulator>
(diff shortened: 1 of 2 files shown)


nvgoldin:
@gbenhaim  - I see your point, though I'm not sure what is the best approach here.
First - as you said, we might have a mixed el6/el7 environment, so  choosing a global templates is probably not a good idea.
Second - "legacy" is ambiguous - we tested that it works with el6, will it work with fc19? I have no idea, and we probably shouldn't test it either, I think that 'legacy' can imply "everything we don't know if its working", while "el6" states explicitly that we keep supporting el6. Further, once we don't support el6(which would happen sometime soon), we can safely drop this template and issue a meaningful message.
Third - I'm not sure if we should allow controlling which templates are generated in the CLI, it makes our options way too verbose, so for instance in the specific failure we have now I would prefer adding ability to control whether the console device should be virtio or 'serial', rather than telling the user  "use this template or that". What I mean is  - we offer "take it or leave it" approach, the defaults, uninteresting options(IMO - such as the console device) just work out of the box without the user intervention. If one would like to try out other configurations - he can override them and take respobsility for that. In that aspect, I think previous discussions of adding 'custom devices' or even overriding the domain XML makes sense.


Anyways - I think its urgent as el6.centos doesn't work properly now, I'll open a ticket to discuss how we advance. I think one path to consider is using a better template mechanism like ``jinja2``.

nvgoldin:
Opened - https://github.com/lago-project/lago/issues/516

nvgoldin:
ci merge please

PR #518, opened 2017-04-24
ovirtlago: change --with-vms timeout in 'lago ovirt start' to 8 minutes

nvgoldin:
ovirtlago: reconfigure timeouts in 'lago ovirt start'

1. Increased default timeout for waiting until the VMs connected
to the Engine report 'up' status to 8 minutes.
2. Added '--vms-timeout' parameter to allow configuring that timeout.
3. Increased 'ovirt-engine'/'vdsmd' service timeout to 3 minutes.
4. Added default timeouts in all relevant functions headers, and set
the timeouts explicitly in 'do_ovirt_start' for visibility.

### DIFF
--- a/ovirtlago/cmd.py
+++ b/ovirtlago/cmd.py
@@ -198,10 +198,18 @@ def do_ovirt_stop_vms(prefix, **kwargs):
 
 
 @cli_plugin(help='Start all VMs that are down')
+@cli_plugin_add_argument(
+    '--vms-timeout',
+    help=('Time to wait until the Engine reports the VMs are up.'),
+    dest='vms_timeout',
+    default=8 * 60,
+    type=int,
+    action='store',
+)
 @in_ovirt_prefix
 @with_logging
-def do_ovirt_start_vms(prefix, **kwargs):
-    prefix.virt_env.engine_vm().start_all_vms()
+def do_ovirt_start_vms(prefix, vms_timeout, **kwargs):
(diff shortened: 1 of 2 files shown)


ovirtlago/cmd.py
```diff
@@ -223,22 +231,30 @@ def do_ovirt_status(prefix, **kwargs):
...
+)
 @in_ovirt_prefix
 @with_logging
-def do_ovirt_start(prefix, with_vms, **kwargs):
+def do_ovirt_start(prefix, with_vms, vms_timeout, **kwargs):
     with LogTask('Starting oVirt environment'):
         prefix.start()
         with LogTask('Waiting for ovirt-engine status'):
-            prefix.virt_env.assert_engine_alive()
+            prefix.virt_env.assert_engine_alive(timeout=2 * 60)
```
gbenhaim: I'm afraid that we will hit this timeout also, what do you think ?

nvgoldin: Could be - this would mean it takes libvirt to boot the VM and until the engine service is running more than 2 minutes, which seems a lot. I never had it timed out on that stage, though I never had it timed out on the 'vms' start either.  I actually think that the timeout for the VMs now is huge - 8 minutes. But practically I can't think of any other 'hot fix' for that..

You experienced it timing out on this engine service stage?

nvgoldin: Updated to 3 minutes.

gbenhaim:
ci merge please

PR #521, opened 2017-04-26
[do not merge] Allow Network Manager to configure interfaces on EL7/F24

mykaul:
If we run a modern guest, we should strive to use Network Manager
to configure interfaces.
This just deletes the configuration of ifcfg-* and allows NM to work.

Tested using ovirt-system-tests on oVirt 4.1 - it almost succeeds.
Need to see what in the test has some assumptions on network interface
names, order, etc.
There's currently an issue with migrations - for some reason, getting no route to host when testing, although a ping later succeeds. @lenoch might have a clue.

### DIFF
--- a/docs/Installation.rst
+++ b/docs/Installation.rst
@@ -2,41 +2,44 @@
 Installing Lago
 ###############
 
-Lago is officially supported and tested on Fedora 24+ and CentOS 7.3
-distributions. However, it should be fairly easy to get it running on
-debian variants.
-
-As Lago requires libvirtd installed and several group permissions,
-it cannot be installed solely via ``pip``. For that reason the
-recommended method of installation is using the RPM. The easiest way, is
-to use the `Install script`_ which we test and verify regularly [2]_.
-
+Lago is officially supported and tested on Fedora 24+ and CentOS 7.3+
+distributions. However, it should be fairly easy to install it on any Linux
+distribution that can run libvirt and qemu-kvm using `pip`_,
+here we provide instructions also for Ubuntu 16.04 which we test from time to
+time. Feel free to open PR if you got it running on a distribution which is not
+listed here so it could be added.
 
 pip
 ===
 
-1. Install system package dependencies, this may vary according to your
-   distribution:
+1. Install system package dependencies:
 
    A. CentOS 7.3+
 
    .. code:: bash
 
         $ yum install -y epel-release centos-release-qemu-ev
-        $ yum install -y libvirt libvirt-devel libguestfs-tools \
-            libguestfs-devel gcc libffi-devel openssl-devel \
-            qemu-kvm-ev
+        $ yum install -y python-devel libvirt libvirt-devel \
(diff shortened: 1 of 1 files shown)


mykaul:
if you have everything set up using ifcfg-eth* files:
1. NetworkManager is not configuring much.
2. You are stuck with 'eth' convention (so I could not test with e1000 for example)


Please note that we should not merge it until we understand why migration fails. It fails on 'no route' though a ping later succeeds - might be some kind of race on setting up the route table and trying to migrate, though I've added a lengthy sleep and it did not help.

lenoch:
I cannot comment on this patch and `(ovirt)lago`(?) tests as I don't use NetworkManager on my system.

Migrations in OST fail (but only with my `remove-vlan-tagging-and-verify-dhcp` topic, am I wrong?) because of two reasons acting at once, but the explanation is simple. Migrations require a working host name resolver because the destination host is identified by name (host0/host1), not just an IP.

1. The three libvirt networks that the hosts are connected to advertise three distinct nameservers over DHCP. As the networks are detached in `003_basic_networking`, the last surviving nameserver is not the one you would think (the management network's) but actually the nameserver belonging to the last detached network (and thus inaccessbile). One way to fix this is by only accepting DNS configuration for the management (default-route) network by setting `PEERDNS=no` in the other networks.
2. NetworkManager gets in the way of VDSM's initscripts-based network setup. The above is true only if it is disabled. If it is running (by default, because for most networking tasks it is avoided by `NM_CONTROLLED=no`), all nameserver configuration (in `/etc/resolv.conf`) is actually wiped out after `003_basic_networking`.

I have a VDSM patch to fix the first issue and I have been writing a non-OST specific explanation of the second issue so I can file a NetworkManager bug.

lenoch:
@mykaul How reproducible are migration failures for you? I believe the current OST code (where my DHCP topic cannot be merged exactly due to migrations failing) is safe and sound.

lenoch:
And before we merge this – Fedora 25 is already out :-)

mykaul:
It's 100 reproducible for me, including all my efforts with time.sleep(60) in between.
I'm using a Fedora 25 host on L0, L1 is of course CentOS 7.3.
I do see migration is trying to use an IP, not hostname though.

nvgoldin:
>The three libvirt networks that the hosts are connected to advertise three distinct nameservers over DHCP. As the networks are detached in 003_basic_networking, the last surviving nameserver is not the one you would think (the management network's) but actually the nameserver belonging to the last detached network (and thus inaccessbile). One way to fix this is by only refusing to accept DNS configuration non-management (non-default-route) networks by setting PEERDNS=no in their ifcfg files.

As for the Lago side of things here - since 0.37 only the management network serves DNS, but it differes in el7/fc25 - In el7, since libvirt 2.2 is not available - the none-mgmt networks use ``dns forward`` to the management network. In @mykaul installation which uses ``fc25`` it should be actually disabling them and not serving DNS at all(assuming libvirt > 2.2).

PR #533, opened 2017-05-07
tests: fix ordering in TestPrefixNetworkInitalization

nvgoldin:
Small issue.
The exception was asserted over 'net-1, net-2,' pattern, while
it could be 'net-2, net-1'. So switching to a regex instead.


nvgoldin:
ci merge please

PR #542, opened 2017-05-17
[WIP] SDK functional tests

nvgoldin:
These tests will be running after the normal CLI functional tests. They use the installed Lago RPMs, but also a venv(which is setup by tox) for the pytest dependencies. 
Currently only basic sanity tests.


nvgoldin:
will go with https://github.com/lago-project/lago/pull/531

Issue #429, opened 2017-01-16
Define ansible role per VM


nvgoldin:
When https://github.com/lago-project/lago/pull/428 gets in, it will deduce the VM ansible role from the vm-type, this might be good for ovirtlago, but would enforce creating a VM plugin per role. I suggest to add to the VM metadata a 'role' parameter, for example:
```yaml
dom1:
....
    metadata:
        role: httpd
```
The ``role`` would default to the VM type if not defined.
I'm familiar enough with ansible requirements, maybe it should be 'ansible' specific? for instance:
```yaml
dom1:
....
    metadata:
    ....
      ansible:
         role: 
         ...
         (other ansible metadata params)
        ...
```
some relevant past issue/work done by  @ehelms : https://github.com/lago-project/lago/issues/281

ifireball:
"role" is something very specific in the Ansible world - A set of tasks and related things that are wrapped together. A role in Ansible is somewhat like a class and a module in Puppet wrapped together into one.

I suggest we find a more generic way to mark up and group VMs, maybe 'labels' or 'tags'.

nvgoldin:
And then parse ``lable`` -> ``role`` in ansible specific commands? Our 'metadata' is pretty context free at the moment. My question is exactly that - will we need more 'ansible' specific paramaters there, then we might bundle them all together.

ifireball:
An Ansible playbook is the way you assign roles to hosts or groups thereof (its more then that, b/c it also specifies the order in which things happen, what can happen in parallel, etc.). This is a core concept of Ansible and Lago shouldn't mess with that too much.

In the end Lago should not be doing a lot more then providing an Inventory file to Ansible. An Inventory file can:
* Define hosts and how to connect to them
* Group hosts
* Define host metadata (AKA variables. At the individual host level or at the group level)

All the above can be useful in the Ansible contest but also in other contexts. consider for example:
* `lago start <name of group of VMs>`
* Making an equvalent of the Ansible commands but for Puppet e.g. generate puppet ENC data where variables become parameter assignments

Bottom line - I thing we should avoid Ansible-specific terminology in the init file and instead come up with concepts that can have wider use.

mykaul:
I'd be happy with Ansible specific terminology. It's a widely-known, used and agreed-upon terminology.

ifireball:
[...]

PR #544, opened 2017-05-21
Refactor lago ansible_hosts

gbenhaim:
1. Every key in the init file can be used as a group.
   In order to create a unique name for each group, the
   value is prefixed with the series of keys and indexes that
   points to it (for example [vm-type=ovirt-host]).

2. Added the option to specify 'groups' key in the domain definition,
   this key should point to a list of groups that the domain belongs to.

3. Moved the logic to a new file called ansible.py (this way 'lago sdk'
   would be able to use it without the need for cmd.py).

For more info check out the description in cmd.py 

closes https://github.com/lago-project/lago/issues/429


lago/plugins/vm.py
```diff
@@ -419,6 +419,18 @@ def mgmt_name(self):
     def mgmt_net(self):
         return self.virt_env.get_net(name=self.mgmt_name)
 
+    @property
+    def vm_type(self):
+        return self._spec['vm-type']
+
+    @property
+    def groups(self):
```
nvgoldin: docstrings

lago/cmd.py
```diff
@@ -492,47 +487,67 @@ def do_console(prefix, host, **kwargs):
...
+    """
+    )
+)
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--keys',
+    '-k',
+    help='Path to the keys that should be used as groups',
+    default=['vm-type', 'groups', 'vm-provider'],
+    metavar='KEY',
+    nargs='*',
```
gbenhaim: It's not.
Currently OST Ansible uses ```[ovirt-host], [ovirt-engine]``` while the new names
are ```[vm-type=ovirt-host], [vm-type=ovirt-engine]```.
Since any key can be used as a group, I have to keep it this way (in order to create a unique group name).

lago/ansible.py
```diff
@@ -0,0 +1,131 @@
+from collections import defaultdict
```
gbenhaim: done

lago/ansible.py
```diff
@@ -0,0 +1,131 @@
...
+        for name, hosts in inventory.viewitems():
+            lines.append('[{name}]'.format(name=name))
+            for host in sorted(hosts):
+                lines.append(host)
+
+        return '\n'.join(lines)
+
+    def get_inventory(self, keys=None):
+        """
+        Create an Ansible inventory based on python object.
```
gbenhaim: done

lago/cmd.py
```diff
@@ -492,47 +487,67 @@ def do_console(prefix, host, **kwargs):
...
+        """
+    This method iterates through all the VMs and creates an Ansible
+    host inventory. For each vm it defines an IP address and a private key.
+
+    The default groups are based on the values which associated
+    with the following keys: 'vm-type', 'groups', 'vm-provider'.
+
+    The 'keys' parameter can be used to override the default groups,
+    for example to create a group which based on a 'service_provider',
+    --keys 'service_provider' should be added to this command.
```
gbenhaim: Thanks :)

lago/plugins/vm.py
gbenhaim: done

gbenhaim:
ci merge please

Issue #516, opened 2017-04-23
Improve templating mechanism


nvgoldin:
When starting a libvirt domain/network, we have 2 stages right now:
1. A central domain template, with place-holders, which we replace prior to parsing the template as XML.
2. XML parts generated dynamically and added to the template(after it was initialized as an ``lxml.Etree``)

In the current state, most logic is shifting towards 2. This gives us more control on the XML, but on the other hand can potentially complicate the code, for instance, https://github.com/lago-project/lago/pull/515 suggests adding a different template for ``el6``. It could be that using a template engine, such as ``Jinja`` could simplify greatly our code, and shift the "boring uninteresting stuff" back to (1), while focusing in (2) on the crucial parts.

For example this might be a solution to #504 (need to define a different console type for el6) using Jinja:
```python
In [1]: from jinja2 import Template

In [2]: domain_template = """
   ...: <domain>
   ...:   ...
   ...:   {% if distro == 'el6' %}
   ...:   <serial type='pty'>
   ...:      <target port='0'>
   ...:   </serial>
   ...:   {% endif %}
   ...:   ...
   ...:   <console type='pty'>
   ...:      <target type={{ 'serial' if distro == 'el6' else 'virtio' }} port='0'/>
   ...:   </console>
   ...:   ...
   ...: </domain>
   ...: """
   ...: 

In [3]: template = Template(domain_template)
   ...: 

In [4]: print(template.render(distro='el6'))
   ...: 

<domain>
  ...
  
  <serial type='pty'>
     <target port='0'>
  </serial>
  
  ...
  <console type='pty'>
     <target type=serial port='0'/>
  </console>
  ...
</domain>

In [5]: print(template.render(distro='el7'))
   ...: 

<domain>
  ...
  
  ...
  <console type='pty'>
     <target type=virtio port='0'/>
  </console>
  ...
</domain>

In [6]: 
```

nvgoldin:
Left to work on the domain templates

PR #547, opened 2017-05-24
 sysprep: support debian network scripts locations

nvgoldin:
1. In debian, the network scripts are found in:
    ``/etc/network/interfaces``
  And per-interfaces configurations can be loaded from:
    ``/etc/network/interfaces.d/``
  In Ubuntu 16.04, the naming scheme changed, (eth -> ens), however, it
  can still be configured by updating the kernel parameters in grub, which is what 
  we'll do for now when building the images. So keeping 'eth' convention for debian. Although
  this is *probably* not the best thing to do(go backwards with the interface names),
  But changing the interfaces names in our sysprep methods would make
  the code more complex, and I rather first implement https://github.com/lago-project/lago/issues/516.
     * Note that this does not allow NetworkManager to configure the
  interfaces, however, Ubuntu 16.04 does not come with it pre-installed with it,
  and installing takes around 250MB.

2. The second commit adds Fedora 25 support, by simply disabling network scripts creation, as NetworkManager is installed and enabled by default and configures all interfaces properly.

### DIFF
--- a/lago/providers/libvirt/vm.py
+++ b/lago/providers/libvirt/vm.py
@@ -193,21 +193,25 @@ class LocalLibvirtVMProvider(vm_plugin.VMProviderPlugin):
                     sysprep.set_iscsi_initiator_name(self.vm.iscsi_name())
                 ]
 
-                if self.vm.distro() in ('el7', 'fc24', 'fc25', 'fc26'):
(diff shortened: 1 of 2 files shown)


nvgoldin:
ci test please

nvgoldin:
ci merge please

PR #554, opened 2017-05-29
 Move sysprep to Jinja2 templates 

nvgoldin:
Shifting the sysprep logic to Jinja2 templates, and calling 'sysprep' with the generated file. This allows splitting the logic to different distros, re-use templates where applicable and most importantly - makes it easier to maintain. 

How it works
-----------------
On bootstrap stage, we try to load ``template/sysprep-{distro}.j2`` template. If it is not available, we load ``templates/sysprep-base.j2``. Each template can include the base one(currently all do).

The structure of templates directory is:
```bash
$ tree templates
templates
├── sysprep-base.j2
├── sysprep-debian.j2
├── sysprep-el6.j2
├── sysprep-el7.j2
├── sysprep-fc24.j2
├── sysprep-fc25.j2
├── sysprep-fc26.j2
└── sysprep-macros.j2
```
Where except the distro specific templates, ``sysprep-macros.j2`` includes some useful Jinja2 macros to make the file generation easier.
So eventually in each distro it should be pretty obvious what is happening when opening the template, for example, this is the ``el7`` one:
```jinja2
{% import 'sysprep-macros.j2' as macros %}
{% include 'sysprep-base.j2' %}

{{ macros.iscsi(name=iscsi_name, hostname=hostname) }}
{{ macros.network_devices_el(mappings=mappings) }}

delete /etc/systemd/system/multi-user.target.wants/kdump.service
delete /etc/systemd/system/multi-user.target.wants/postfix.service
```

Few other stuff:
1. I dropped some commands in sysprep.py(such as --update), because we never called them, and they can be easily append to the template. 
2. Future work could include default template per distro(without version) - such as Fedora/EL/Debian.

### DIFF
--- a/automation/check-patch.packages
+++ b/automation/check-patch.packages
@@ -6,6 +6,7 @@ git
 grubby
 libffi-devel
 libvirt-devel
+libguestfs-devel
 libxslt-devel
 openssl-devel
 pandoc
--- a/automation/check-patch.packages.el7
+++ b/automation/check-patch.packages.el7
@@ -3,6 +3,7 @@ bats
 git
 libffi-devel
 libvirt-devel
+libguestfs-devel
 libxslt-devel
 openssl-devel
 pandoc
--- a/lago.spec.in
+++ b/lago.spec.in
@@ -64,7 +64,9 @@ BuildRequires: python-wrapt
 %if 0%{?fedora} >= 24
 BuildRequires: python2-configparser
 BuildRequires: python2-paramiko >= 2.1.1
+BuildRequires: python2-jinja2
 %else
+BuildRequires: python-jinja2
 BuildRequires: python-configparser
 BuildRequires: python2-paramiko
 %endif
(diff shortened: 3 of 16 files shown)


nvgoldin:
wops, forgot the grub update, fixing.

nvgoldin:
done

mykaul:
Very interesting. Brings us closer to Ansible, as this template probably can be used when/if we'll call Ansible to do the provisioning. Would be great to move the dom_template.xml to a Jinja2 template (example form Linchpin - https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/provision/roles/libvirt/templates/libvirt_node.xml.j2 )

LGTM, but need to compare the result virt-sysprep command line to ensure nothing was left out.

nvgoldin:
> LGTM, but need to compare the result virt-sysprep command line to ensure nothing was left out.

I'll compare it again before merging - also with OST.

Domain work - https://github.com/lago-project/lago/pull/576, currently only simple stuff, I'll have a look at the LinchPin templates for inspiration. In XML terms, not sure what would be easier to maintain(the xml etree package is pretty good in a way).

nvgoldin:
Tested basic suite master/4.1/image-ng - looks good, except minor issue which is fixed in https://gerrit.ovirt.org/#/c/78376/3.

nvgoldin:
ci merge please

nvgoldin:
ci merge please

nvgoldin:
ci merge please

nvgoldin:
ci merge please

nvgoldin:
ci merge please

PR #555, opened 2017-05-30
sysprep: ensure el based ifcfg files are created consistently

nvgoldin:
Using the SDK I see random failures on boot to acquire IP via DHCP,
looking at the logs, it looks like the order of keys and values in the
generated ifcfg-eth* devices is different than successful runs in the
CLI. Also I see failures when the 'TYPE' parameter is first, while
successes when 'HWADDR' is first. This in all might be vodoo tales, but
it might solve this issue, before digging deeper into what is
wrong with NetworkManager(the logs don't reveal much currently - just
that dhclient start, but it doesn't send any requests).

If this is only vodoo indeed, this patch can't harm(just ensures we use
the same order).

### DIFF
--- a/lago/sysprep.py
+++ b/lago/sysprep.py
@@ -21,6 +21,9 @@ import os
 
 import utils
 from textwrap import dedent
+import tempfile
+import logging
+LOGGER = logging.getLogger(__name__)
 _DOT_SSH = '/root/.ssh'
 _AUTHORIZED_KEYS = os.path.join(_DOT_SSH, 'authorized_keys')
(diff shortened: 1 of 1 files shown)


nvgoldin:
ci test please

nvgoldin:
ci test please

nvgoldin:
Tested on OST(basic-suite/he 4.1) and locally on mutli-distro setup, no failures noticed, so merging.
It might have been escaping in the file also(we didn't have '\n' in the last line).

nvgoldin:
ci merge please

PR #565, opened 2017-06-11
Add the '--delete' to sysprep so we can delete files during sysprep.

mykaul:
It is used to disable services to boot faster: kdump and postfix.
Specifically disabling kdump reduces boot time nicely.

### DIFF
--- a/lago/providers/libvirt/vm.py
+++ b/lago/providers/libvirt/vm.py
@@ -39,6 +39,9 @@ LOGGER = logging.getLogger(__name__)
 LogTask = functools.partial(log_utils.LogTask, logger=LOGGER)
 log_task = functools.partial(log_utils.log_task, logger=LOGGER)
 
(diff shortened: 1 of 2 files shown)


nvgoldin:
1. @mykaul  - my bad, missed this. Can you rebase?
2. In the context of OST - I wonder if the vdsm sos plugin uses kdump.
3. I triggered the [check-merged](http://jenkins.ovirt.org/job/lago_master_github_check-merged-el7-x86_64/100) job which passed good, though sort of hard to see if it actually improved the boot time in CI.

mykaul:
We don't use the SOS plugin - and log collector seems to work. kdump can be optionally used in case we fence - and we don't in OST.

It's easy to see how it improves boot time if you run 'systemd-analyze blame' before and after. On my laptop it reduces ~10 seconds.

nvgoldin:
ci merge please

PR #589, opened 2017-06-23
Fix SSH for VMs without public key

mykaul:
This seems to fix the usage of SSH for VMs without public key.
On the way, it simplifies the code a bit.


gbenhaim:
ci merge please

PR #600, opened 2017-06-25
automation: merge check-patch and check-merged

nvgoldin:
1. Use 'run_tests.sh' to trigger both easily.
2. Fail early on each stage(don't keep running if any stage failed)
3. Remove 'exported-artifacts' directory on first run
4. If changing only the ``docs`` directory - don't run any tests, just build the docs(both check-patch and merged).


nvgoldin:
ci merge please

PR #609, opened 2017-06-27
Fixing export issues

gbenhaim:
1. Allow specifying in the init file which vms / disks to export.
2. Don't run virt-sparsify on iso disks.
3. Use deepcopy when generating the env's spec (will help
   to avoid unwanted changes to the internal metadata of the env).
4. Adding collect-only: only show which disks are going to be exported.
5. Expose export to lago.sdk
6. Safely try to parse the name of the backing image

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -344,15 +344,23 @@ def do_shutdown(prefix, vm_names, reboot, **kwargs):
     default='LagoInitFile',
     help='The name of the exported init file',
 )
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--collect-only',
+    action='store_true',
+    help='Only output the disks that will be exported',
+)
 @in_lago_prefix
 @with_logging
 def do_export(
     prefix, vm_names, standalone, dst_dir, compress, init_file_name,
-    out_format, **kwargs
+    out_format, collect_only, **kwargs
 ):
-    prefix.export_vms(
-        vm_names, standalone, dst_dir, compress, init_file_name, out_format
+    output = prefix.export_vms(
+        vm_names, standalone, dst_dir, compress, init_file_name, out_format,
+        collect_only
     )
+    if collect_only:
+        print(out_format.format(output))
 
 
 @lago.plugins.cli.cli_plugin(
--- a/lago/export.py
+++ b/lago/export.py
@@ -185,9 +185,16 @@ class TemplateExportManager(DiskExportManager):
                 # this identifier will be used later by Lago in order
                 # to resolve and download the base image
                 parent = self.src_qemu_info[0]['backing-filename']
-                parent = './{}'.format(
-                    os.path.basename(parent.split(':', 1)[1])
-                )
+                # Hack for working with lago images naming convention
+                # For example: /var/lib/lago/store/phx_repo:el7.3-base:v1
+                # Extract only the image name (in the example el7.3-base)
+                parent = os.path.basename(parent)
+                try:
+                    parent = parent.split(':', 1)[1]
+                except IndexError:
+                    pass
+
+                parent = './{}'.format(parent)
                 utils.qemu_rebase(
                     target=self.dst, backing_file=parent, safe=False
                 )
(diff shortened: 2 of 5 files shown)


lago/providers/libvirt/vm.py
```diff
@@ -320,8 +319,7 @@ def export_disks(self, standalone, dst_dir, compress, *args, **kwargs):
                 standalone=standalone,
                 *args,
                 **kwargs
-            ) for disk in self.vm.disks
-            if disk.get('format') not in formats_to_exclude
+            ) for disk in self.vm.disks if not disk.get('skip-export')
```
nvgoldin: Please add it to the LagoInitFile docs, small issue but would make our life easier.

lago/virt.py
```diff
@@ -210,6 +214,8 @@ def generate_init(self, dst, out_format, filters=None):
                 formatter for the output (the default is yaml)
             filters (list): list of paths to keys that should be removed from
                 the init file
+            vms_to_include (list of lago.plugins.vm.VMPlugin): list of vms to
```
nvgoldin: need :class:`lago.plugins.vm.VMPlugin` here

lago/cmd.py
```diff
@@ -344,15 +344,23 @@ def do_shutdown(prefix, vm_names, reboot, **kwargs):
     default='LagoInitFile',
     help='The name of the exported init file',
 )
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--collect-only',
```
nvgoldin: nice

lago/prefix.py
```diff
@@ -1220,16 +1221,35 @@ def build(self, conf):
...
+                export all the vms in the env
+            standalone(bool): If false, export a layered image
+            export_dir(str): Dir to place the exported images and init file
+            compress(bool): If True compress the images with xz
+            init_file_name(str): The name of the exported init file
+            out_format(lago.plugins.output.OutFormatPlugin):
+                The type of the exported init file (the default is yaml)
+            collect_only(bool): If true, return only a mapping from vm name
+                to the disks that will be exported.
+
```
nvgoldin: What does it return in the general case?
(None -> if not collect_only?)

lago/export.py
```diff
@@ -185,9 +185,16 @@ def rebase(self):
...
                 parent = self.src_qemu_info[0]['backing-filename']
-                parent = './{}'.format(
-                    os.path.basename(parent.split(':', 1)[1])
-                )
+                # Hack for working with lago images naming convention
+                # For example: /var/lib/lago/store/phx_repo:el7.3-base:v1
+                # Extract only the image name (in the example el7.3-base)
+                parent = os.path.basename(parent)
+                try:
+                    parent = parent.split(':', 1)[1]
```
nvgoldin: Although a nasty hack, i think adding a regex matching here would make the hack "nicer" and cleaner.

lago/virt.py
gbenhaim: done

lago/cmd.py
gbenhaim: thanks !

lago/prefix.py
gbenhaim: yes.
Fixed the docstrings.

lago/export.py
gbenhaim: done

gbenhaim:
Thanks for the review.
I will send the updated patch tomorrow.

lago/prefix.py
nvgoldin: (again github messed the comments so I can't see the change) - Thinking about it - it might be useful to always return some metadata of the export, including some finalized data(such as where is the .tar.gz):
For example, (semi-pesudoish here):
```python
     export_results = sdk.export(....)
     compressed_export = export_results.get('compressed_file')
     os.cp(compressed_export, os.path.join('exported-artifacts', 'export'))
```

gbenhaim:
Opening a new pull request with the fixes.

PR #610, opened 2017-06-27
Fix sdist creation

nvgoldin:
1. First build the sdist to exported-artifacts.
2. Then build the RPMs(which also create an sdist - but manipulate it).
3. Add installation test from the sdist we exported(and don't let tox
produce it by itself, because that is not the one we publish).
4. Lastly, don't publish the 'tar.gz' the RPMs used, only the sdist.

This avoids the corruption in the sdist that our RPM related scripts
caused(in Makefile). So we ensure the 'sdist' we publish is the one
produced by 'python setup.py sdist'.


nvgoldin:
ci merge please

PR #615, opened 2017-06-28
export: Some fixes

gbenhaim:
1. Allow to specify in the init file which vms / disks to export.

2. Don't run virt-sparsify on iso disks.

3. Use deepcopy when generating the env's spec (will help
   to avoid unwanted changes to the internal metadata of the env).

4. Adding collect only option to export cmd: This option will output
   the mapping from vms to the disks that will be exported.

5. Expose export to lago.sdk

6. Safely try to parse the name of the backing image:
   Due to lago image store naming convention (repo:image:version),
   if the backing image is a Lago image, preprocessing to the name
   needs to be done (in order to extract only the image name and the
   version) and then set it as the backing image of the exported image.

   Since not always the backing image will be a Lago images, adding
   exception handling to the parsing stage.

7. Adding VMExportManager: Becuase of the functionality that was added
   (as described above), the code became complex, so adding
   VMExportManager as an abstraction.

8. Adding the ability to choose if to use threads.

9. prefix.export_vms will return a dict that maps between
   vms names and the path of the disks that were exported.

10. Return threads results from utils.invoke_in_parallel and
    utils.invoke_different_funcs_in_parallel.

### DIFF
--- a/lago/export.py
+++ b/lago/export.py
@@ -16,11 +16,11 @@ LogTask = functools.partial(log_utils.LogTask, logger=LOGGER)
 
 class DiskExportManager(object):
     """
-    ExportManager object is responsible on the export process of an image
-    from the current Lago prefix.
+    DiskExportManager object is responsible on the export process of
+    an image from the current Lago prefix.
 
-    DiskExportManger is the base class of specific ExportManagers.
-    Each specific ExportManger is responsible on the export process of an
+    DiskExportManger is the base class of specific DiskExportManger.
+    Each specific DiskExportManger is responsible on the export process of an
     image with a specific type (e.g template, file...)
 
     Attributes:
@@ -32,7 +32,8 @@ class DiskExportManager(object):
             as found in workdir/current/virt/VM-NAME
         exported_metadata(dict): A copy of the source disk metadata, this
             dict should be updated with new values during the export process.
-            do_compress(bool): If true, apply compression to the exported disk.
+        do_compress(bool): If true, apply compression to the exported
+            disk.
     """
 
     __metaclass__ = ABCMeta
@@ -65,7 +66,8 @@ class DiskExportManager(object):
             disk (dict): Disk attributes
                 (of the disk that should be exported) as
                 found in workdir/current/virt/VM-NAME
-            do_compress(bool): If true, apply compression to the exported disk.
+            do_compress(bool): If true, apply compression to the exported
+                disk.
         Returns
             An instance of a subclass of DiskExportManager which
             matches the disk type.
@@ -185,9 +187,18 @@ class TemplateExportManager(DiskExportManager):
                 # this identifier will be used later by Lago in order
                 # to resolve and download the base image
                 parent = self.src_qemu_info[0]['backing-filename']
-                parent = './{}'.format(
-                    os.path.basename(parent.split(':', 1)[1])
-                )
+
+                # Hack for working with lago images naming convention
+                # For example: /var/lib/lago/store/phx_repo:el7.3-base:v1
+                # Extract only the image name and the version
+                # (in the example el7.3-base:v1)
+                parent = os.path.basename(parent)
+                try:
+                    parent = parent.split(':', 1)[1]
+                except IndexError:
+                    pass
+
+                parent = './{}'.format(parent)
                 utils.qemu_rebase(
                     target=self.dst, backing_file=parent, safe=False
                 )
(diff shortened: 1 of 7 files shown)


gbenhaim:
About the issue with parsing the image name: In addition to the name, we need to save the version of the base image (that is because of the way we query the template store for images), so working with ``str.split`` making sense.

lago/export.py
```diff
@@ -16,11 +16,12 @@
 
 class DiskExportManager(object):
     """
-    ExportManager object is responsible on the export process of an image
-    from the current Lago prefix.
+    :class:DiskExportManager object is responsible on the export process of
```
nvgoldin: Not sure if you need the :class: here, it is meant for cross-references(your already inside the class), also if it is needed you need '`' for sphinx to render it.

lago/export.py
```diff
@@ -32,7 +33,8 @@ class DiskExportManager(object):
             as found in workdir/current/virt/VM-NAME
         exported_metadata(dict): A copy of the source disk metadata, this
             dict should be updated with new values during the export process.
-            do_compress(bool): If true, apply compression to the exported disk.
+            do_compress(bool): If true, apply compression to the exported
```
nvgoldin: over indenteded

lago/export.py
```diff
@@ -145,7 +148,8 @@ def write_lago_metadata(self):
 
 class TemplateExportManager(DiskExportManager):
     """
-    TemplateExportManager is responsible exporting images of type template.
+    :class:TemplateExportManager is responsible exporting images of type
```
nvgoldin: same about the :class:, though might be wrong better check how it rendered them.
(side note - it is a bit confusing as we are using the google style of code, (i.e. Args:/Returns: instead of :class: etc), but for sphinx to render properly you need the :thing:).

lago/export.py
```diff
@@ -185,9 +189,18 @@ def rebase(self):
...
-                    os.path.basename(parent.split(':', 1)[1])
-                )
+
+                # Hack for working with lago images naming convention
+                # For example: /var/lib/lago/store/phx_repo:el7.3-base:v1
+                # Extract only the image name and the version
+                # (in the example el7.3-base:v1)
+                parent = os.path.basename(parent)
+                try:
+                    parent = parent.split(':', 1)[1]
```
nvgoldin: OK - I guess both are viable. personally I  prefer using regex named groups that give you a clue what you extracted from the string.

lago/export.py
```diff
@@ -243,14 +257,98 @@ def export(self):
...
+class VMExportManager(object):
+    """
+    :class:VMExportManager object is responsible on the export process of a
+    list of disks.
+
+    Attributes:
+        disks (list of dicts): Disks to export.
+        dst (str): Where to place the exported disks.
+        compress(bool): If True compress each exported disk.
+        with_threads(bool): If True, run the export in parallel
```
nvgoldin: Maybe indicate that it will run a thread per disk. Also, if its not complicated might change it to an int so that:
threads=0 - > run a thread for each disk
threads=N -> run N threads.

Can be done later of course.

lago/export.py
```diff
@@ -243,14 +257,98 @@ def export(self):
...
+        """
+        return [
+            os.path.join(self._dst, os.path.basename(disk['path']))
+            for disk in self._collect()
+        ]
+
+    def _get_export_mgr(self):
+        """
+        Returns:
+            (:class:DiskExportManager): Handler for each disk
```
nvgoldin: Missing a '`' I think.

lago/plugins/vm.py
```diff
@@ -374,13 +374,21 @@ def extract_paths(self, paths, *args, **kwargs):
         return self.provider.extract_paths(paths, *args, **kwargs)
 
     def export_disks(
-        self, standalone=True, dst_dir=None, compress=False, *args, **kwargs
+        self,
```
nvgoldin: Not per this PR - but the docstring is still a big problem for us, as what you'll see in the SDK is "Thin method that just uses the provider". Guess we could write some wrapper to update the docstring from the service provider dynamically.

lago/utils.py
```diff
@@ -97,13 +97,13 @@ def join_all(self, raise_exceptions=True):
 def invoke_in_parallel(func, *args_sequences):
     vt = VectorThread(func_vector(func, zip(*args_sequences)))
     vt.start_all()
-    vt.join_all()
+    return vt.join_all()
```
nvgoldin: just wondering - why is this change needed?

need to be careful here - OST uses this directly I think.

lago/export.py
gbenhaim: Agree.
It will require a change to ``VectorThread``: https://github.com/lago-project/lago/issues/616

lago/export.py
gbenhaim: done

lago/export.py
gbenhaim: done

lago/export.py
gbenhaim: done

lago/utils.py
gbenhaim: I'm using it in ``virt.py`` line 217, when returning the value from the thread of each vm.
I will test it using the manual job.

gbenhaim:
ci test please

gbenhaim:
verified
http://jenkins.ovirt.org/view/oVirt%20system%20tests/job/ovirt-system-tests_manual/724/

gbenhaim:
ci merge please

gbenhaim:
ci merge please

PR #617, opened 2017-06-29
Get libvirt version correctly

nvgoldin:
In f8be8f4d892149669c9f1abc4bdaf67cc9a5ae2c I misused 'getVersion'
instead of 'getLibVersion'. The first returns the hypervisor version
which was not the intention.

Tested locally where it matters: started with lago a Ubuntu 16.04 VM -> installed Lago from pip -> started a el7.3 nested VM.

### DIFF
--- a/lago/providers/libvirt/vm.py
+++ b/lago/providers/libvirt/vm.py
@@ -57,7 +57,7 @@ class LocalLibvirtVMProvider(vm_plugin.VMProviderPlugin):
             name=self.vm.virt_env.uuid + libvirt_url,
             libvirt_url=libvirt_url,
         )
-        self._libvirt_ver = self.libvirt_con.getVersion()
+        self._libvirt_ver = self.libvirt_con.getLibVersion()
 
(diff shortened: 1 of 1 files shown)


nvgoldin:
BTW , found in libvirt changelog since when you can use ``/dev/urandom``:

>This backend type expects a non-blocking character device as input. The file name is specified as contents of the backend element. Since 1.3.4 any path is accepted. Before that /dev/random and /dev/hwrng were the only accepted paths. When no file name is specified the hypervisor default is used. For qemu, the default is /dev/random 

https://libvirt.org/formatdomain.html

Right now we use ``/dev/urandom`` if libvirt version is >= 2.2 - 
https://github.com/lago-project/lago/blob/master/lago/providers/libvirt/templates/dom_template-base.xml.j2#L30
So theoretically we could lower it to 1.3.4, but as this is only an enhancement I'll keep it this way(unless some unexpected demand comes..).

nvgoldin:
ci merge please

PR #621, opened 2017-07-01
[WIP] Solve some concurrency issues

nvgoldin:
1. Use uuid4 instead of uuid1, so we do not give the same network names in parallel runs.
2. Let libvirt generate the bridge name - here I think I hit occasionally a race condition in libvirt(one run complains the bridge name already exists, because the other process just created it). According to libvirt docs the bridge name will be auto-generated, however it does not talk directly about parallel-safety. If indeed this an issue, we'll have to generate the bridge name too, which will be little more constrained than the network names because it has a lower limitation on the number of chars(16 iirc).


lago/virt.py
```diff
@@ -126,10 +126,15 @@ def prefixed_name(self, unprefixed_name, max_length=0):
                 will adapt the given name and the length of the uuid ot fit it
 
         Returns:
-            str: prefixed identifier for the given unprefixed name
+            str: prefixed identifier for the given unprefixed name, with all
+                none alphanumeric characters in the unprefixed name replaced
+                with a 'L'
```
nvgoldin: Note to self - probably better replacing it with a random char.

nvgoldin:
ci test please

PR #622, opened 2017-07-01
spec: force libguestfs >= 1.30

nvgoldin:
Older versions don't have '--commands-from-file' which we use now
by default.


nvgoldin:
ci merge please

PR #623, opened 2017-07-02
export: Backing file resolution fixes

gbenhaim:
1. If the disk is a layer and the path to his backing
   file is correct, do nothing.

2. Avoid endless recursion when the disk and its backing file
   has the same name.

3. Safely parse the name and the version of the backing file.
   Raise exception if the format of the backing file is not name:version.


gbenhaim:
ci merge please

Issue #620, opened 2017-06-30
TypeError when running `lago list`


d3dave:
```
$ lago list prefixes
Error occured, aborting
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/lago/cmd.py", line 941, in main
    cli_plugins[args.verb].do_run(args)
  File "/usr/local/lib/python2.7/dist-packages/lago/plugins/cli.py", line 184, in do_run
    self._do_run(**vars(args))
  File "/usr/local/lib/python2.7/dist-packages/lago/utils.py", line 501, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/lago/utils.py", line 512, in wrapper
    return func(*args, prefix=prefix, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/lago/cmd.py", line 631, in do_list
    workdir.load()
  File "/usr/local/lib/python2.7/dist-packages/lago/workdir.py", line 158, in load
    basepath, dirs, _ = os.walk(self.path).next()
  File "/usr/lib/python2.7/os.py", line 278, in walk
    names = listdir(top)
TypeError: coercing to Unicode: need string or buffer, NoneType found
```
Same with `list envs`...

nvgoldin:
Thanks for reporting this. I think we already had a ticket about that, but can't find it at the moment. From what I recall, we talked about deprecating the ``lago list`` command or replacing it with something more flexible. @gbenhaim - what do you think?

gbenhaim:
Agree, ```lago list``` should be removed.

gbenhaim:
https://github.com/lago-project/lago/pull/628

gbenhaim:
fixed https://github.com/lago-project/lago/pull/628

PR #628, opened 2017-07-03
Fix lago list

gbenhaim:
1. Fixed lago list to print the available prefixes in a workidir.

2. Fixed the docstrings.

3. Added cli functional test.

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -619,30 +619,26 @@ def do_status(prefix, out_format, **kwargs):
     print(out_format.format(info_dict))
 
 
-@lago.plugins.cli.cli_plugin(
-    help='List the name of the available given virtual resources'
-)
+@lago.plugins.cli.cli_plugin(help='List the prefixes (envs) in a Workdir')
 @lago.plugins.cli.cli_plugin_add_argument(
-    'resource_type',
-    help='Type of resource to list',
-    metavar='RESOURCE_TYPE',
-    choices=['envs', 'prefixes'],
+    'workdir_path',
+    help=dedent(
+        """
+        Path to the Workdir. If not provided Lago will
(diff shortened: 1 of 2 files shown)


gbenhaim:
1. After looking at the code of `lago list` I've realized that its purpose is to list all the prefixes in a workdir,   so fixing the code and the help message.

2. Personally, I don't see any benifit to group prefixes in a workdir (as long as they don't share the  
   same files).

3. Yet, I think that working with multiple prefixes is usefull, and that we should provide a new command,   
     `lago workon` to switch between them (similar to virtualenvwrapper workon).

4. In addition to 3, It will be nice to issue commands to the current env (that was specified with `lago  
   workon`) regardless of the current directory.

5. It will be nice if a user could list all of his existing envs + some description on those envs (as 
    specified in the init file or to `lago init`)

gbenhaim:
closes https://github.com/lago-project/lago/issues/620

nvgoldin:
> Yet, I think that working with multiple prefixes is usefull, and that we should provide a new command,
lago workon to switch between them (similar to virtualenvwrapper workon).

That is what ``lago set-current`` is supposed to do I think. Or you mean *globally*?

gbenhaim:
globally !

gbenhaim:
ci test please

nvgoldin:
``venvwrapper`` is a little different because it assumes you put all your venvs in the same directory, and then you run ``export WORK_ON=...``.

gbenhaim:
Correct, but you are talking about implementation, and I've just mentioned the required (IMO) functionality.

gbenhaim:
ci test please

gbenhaim:
ci test please

gbenhaim:
ci merge please

Issue #607, opened 2017-06-27
Should we stop lago if plugin loading failed ?


gbenhaim:
Currently, if loading a plugin fails, we just log it. I'm not sure if this is a good idea or not.
what do you think? 

```
2017-06-27 12:21:07,024::__init__.py::<lambda>::68::lago.plugins::DEBUG::Could not load 'ovirt-he-host': No module named ovirtsdk.api
2017-06-27 12:21:07,024::__init__.py::<lambda>::68::lago.plugins::DEBUG::Could not load 'ovirt-node': No module named ovirtsdk.api
2017-06-27 12:21:07,024::__init__.py::<lambda>::68::lago.plugins::DEBUG::Could not load 'ovirt-host': No module named ovirtsdk.api
```

gbenhaim:
I think that a proper solution will be to print a warning, instead of just dumping the error to the log.

PR #635, opened 2017-07-09
Print warning when failed to load plugin

gbenhaim:
closes https://github.com/lago-project/lago/issues/607

### DIFF
--- a/lago/plugins/__init__.py
+++ b/lago/plugins/__init__.py
@@ -65,7 +65,7 @@ def load_plugins(namespace, instantiate=True):
         namespace=namespace,
         on_load_failure_callback=(
             lambda _, ep, err: LOGGER.
-            debug('Could not load %r: %s', ep.name, err)
+            warning('Could not load plugin {}: {}'.format(ep.name, err))
         )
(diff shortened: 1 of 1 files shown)


lago/plugins/__init__.py
```diff
@@ -64,8 +65,8 @@ def load_plugins(namespace, instantiate=True):
     mgr = ExtensionManager(
         namespace=namespace,
         on_load_failure_callback=(
-            lambda _, ep, err: LOGGER.
-            debug('Could not load %r: %s', ep.name, err)
+            lambda _, ep, err: warnings.
+            warn('Could not load plugin {}: {}'.format(ep.name, err))
```
nvgoldin: I think using ``LOGGER.warning`` would be better here, iirc it would also go both to stderr and to the log file.
I wanted to use it also in the ``lago/__init__.py`` guestfs warning, but at that stage, logging is not setup yet.

gbenhaim: done

gbenhaim:
ci merge please

Issue #614, opened 2017-06-28
The subnet is already taken by /tmp/pytest-of-root/pytest-21/env0/lago/default/uuid


nvgoldin:
@gbenhaim 
http://jenkins.ovirt.org/job/lago_master_github_check-patch-fc25-x86_64/450/testReport/junit/tests.functional-sdk/test_sdk_sanity/test_custom_log/

I am actually seeing this locally quite often since https://github.com/lago-project/lago/commit/0c8de17447683c1f3ca8da02929d9f27b13cc099 was merged. Perhaps now the code does what it was supposed to do?

I think what happened per the job that failed - init failed in a previous job, but the ``/tmp/....`` lago environment was not destroyed and the subnet lease remained in the locks.

nvgoldin:
Looks like we're partly affected by https://github.com/pytest-dev/pytest/issues/1120
Regardless, need to think of a proper way to clean it up.

nvgoldin:
Another failure: http://jenkins.ovirt.org/job/lago_master_github_check-patch-fc25-x86_64/451/testReport/tests.functional-sdk/test_sdk_sanity/test_custom_log/
cleaned it up manually on the slave

gbenhaim:
In another build, I saw that the networks weren't destroyed.
Taken from the slave: 

```
virsh # net-list
 Name                 State      Autostart     Persistent
----------------------------------------------------------
 default              active     yes           yes
 dfe5-net-01          active     no            no
 dfe5-net-02          active     no            no
```

nvgoldin:
which build?

gbenhaim:
http://jenkins.ovirt.org/job/lago_master_github_check-patch-fc25-x86_64/457/

nvgoldin:
Looking at the slave's build history, these might be leftovers from:
http://jenkins.ovirt.org/job/lago_master_github_check-patch-fc25-x86_64/449/console

Where the job simply timed out:
```
17:16:40 test_sdk_sanity.py::test_extract_paths_ignore_nopath[vm-el7-3-base-/nothing/here-normal] PASSED
22:52:26 test_sdk_sanity.py::test_extract_paths_ignore_nopath[vm-el7-3-base-/nothing/here-dead] Build timed out (after 360 minutes). Marking the build as failed.
22:52:26 $ ssh-agent -k
22:52:26 unset SSH_AUTH_SOCK;
```
[...]

Issue #599, opened 2017-06-25
subnets: free leases on destroy


nvgoldin:
@gbenhaim  - I have second thoughts about:
https://github.com/lago-project/lago/blob/master/lago/subnet_lease.py#L59-L61
```
  Cleanup of stale leases is done in a lazy manner during a request for a
    lease. The store will remove at most 1 stale lease in each request (see
    SubnetStore._lease_valid for more info).
```

Why don't we release the subnet in the prefix destroy method? It only makes sense that when you destroy an environment you would free the subnet, instead of waiting for a different invocation that would clean it up for you. 
The "lazy manner" removal can still be kept, but it is less reliable(as you might not have access to the environment to know if it still exists).

Missed something here?

gbenhaim:
Agree, we should add subnet cleanup when destroying an env.

PR #637, opened 2017-07-11
subnets: free on destroy

nvgoldin:
Free the used subnet when destroying the prefix.

fixes: https://github.com/lago-project/lago/issues/614
fixes: https://github.com/lago-project/lago/issues/599

### DIFF
--- a/lago/subnet_lease.py
+++ b/lago/subnet_lease.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2014 Red Hat, Inc.
+# Copyright 2014-2017 Red Hat, Inc.
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -17,6 +17,8 @@
 #
 # Refer to the README and COPYING files for full details of the license
 #
+
+from future.utils import raise_from
 import functools
 import json
 import os
@@ -106,12 +108,13 @@ class SubnetStore(object):
             otherwise create it.
         """
         try:
-            if not os.path.isdir:
+            if not os.path.isdir(self.path):
                 os.makedirs(self.path)
         except OSError as e:
-            # errno 13 = Permission denied
-            if e.errno == 13:
-                raise LagoSubnetLeaseBadPermissionsException(self.path)
+            raise_from(
+                LagoSubnetLeaseBadPermissionsException(self.path, e.strerror),
+                e
+            )
 
     def acquire(self, uuid_path, subnet=None):
         """
@@ -207,9 +210,8 @@ class SubnetStore(object):
         if not lease.exist:
             return None
 
-        env = lease.has_env
-        if env:
-            return env
+        if lease.has_env:
+            return lease.uuid_path
         else:
             self._release(lease)
             return None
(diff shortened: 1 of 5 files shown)


tests/unit/lago/test_subnet_lease.py
```diff
@@ -117,3 +128,42 @@ def test_take_stale_lease(self, subnet_store, prefixes):
...
+    @pytest.mark.parametrize('remains', [0, 3, 6])
+    def test_release_several_prefixes(
+        self, subnet_store, prefix_mock_gen, num_prefixes, remains
+    ):
+        gen = prefix_mock_gen()
+        for _ in range(num_prefixes):
+            subnet_store.acquire(next(gen).uuid_path)
+        acquired = subnet_store.list_leases()
+        assert all(lease.valid for lease in acquired)
+        assert len(subnet_store.list_leases()) == num_prefixes
```
gbenhaim: s/subnet_store.list_leases()/acquired

gbenhaim: or even:
`assert acquired == num_prefixes == len(os.listdir(subnet_store.path))`
and remove line 143

tests/unit/lago/test_subnet_lease.py
```diff
@@ -117,3 +128,42 @@ def test_take_stale_lease(self, subnet_store, prefixes):
...
+        assert len(os.listdir(subnet_store.path)) == num_prefixes
+        to_release = random.sample(acquired, num_prefixes - remains)
```
gbenhaim: Nice

tests/unit/lago/test_subnet_lease.py
```diff
@@ -117,3 +128,42 @@ def test_take_stale_lease(self, subnet_store, prefixes):
...
+        subnet_store.release([lease.subnet for lease in to_release])
+        assert len(subnet_store.list_leases()) == remains
```
gbenhaim: Same here, lines 146 and 147 can be joined into one line.

tests/unit/lago/test_subnet_lease.py
```diff
@@ -117,3 +128,42 @@ def test_take_stale_lease(self, subnet_store, prefixes):
...
+        assert len(os.listdir(subnet_store.path)) == remains
+
+    def test_list_leases_raises(self, tmpdir):
+        with patch('lago.subnet_lease.os.listdir') as mock_listdir:
+            subnet_store = subnet_lease.SubnetStore(path=str(tmpdir))
```
gbenhaim: Why not use `subnet_store` fixture?

lago/subnet_lease.py
```diff
@@ -245,26 +244,53 @@ def _take_lease(self, lease, uuid_path, safe=True):
             'Assigned subnet lease {} to {}'.format(lease.path, uuid_path)
         )
 
-    def release(self, subnet):
+    def list_leases(self, uuid=None):
```
nvgoldin: Will add.

tests/unit/lago/test_subnet_lease.py
nvgoldin: +1, will change.

lago/subnet_lease.py
```diff
@@ -245,26 +244,53 @@ def _take_lease(self, lease, uuid_path, safe=True):
...
+            lease_files = os.listdir(self._path)
+        except OSError as e:
+            # errno 13 = Permission denied
+            if e.errno == 13:
+                raise LagoSubnetLeaseBadPermissionsException(self.path)
+            # errno 2 = No such file or directory
+            if e.errno == 2:
+                return []
+            else:
+                raise
```
nvgoldin: Will update. Thinking about it, better not swallow the `e.errno == 2` case, I'll just raise there too.

lago/subnet_lease.py
```diff
@@ -245,26 +244,53 @@ def _take_lease(self, lease, uuid_path, safe=True):
             'Assigned subnet lease {} to {}'.format(lease.path, uuid_path)
         )
 
-    def release(self, subnet):
+    def list_leases(self, uuid=None):
+        try:
+            lease_files = os.listdir(self._path)
```
nvgoldin: Will change.

lago/subnet_lease.py
```diff
@@ -207,9 +207,8 @@ def _lease_valid(self, lease):
         if not lease.exist:
             return None
 
-        env = lease.has_env
-        if env:
-            return env
+        if lease.has_env:
```
nvgoldin: Not what I had in mind - in `Lease` class I changed 'has_env' to return a Boolean variable, which I think is what you'd expect from its name(has_env).
I think the performance issue here is probably neglectable, but I could:
change `has_env -> get_env` - and make it return the env or `None`, what do you think?

tests/unit/lago/test_subnet_lease.py
nvgoldin: Thanks, missed that.
About the triple '==' comparison, it might make the code more compact, but not pytest's output :)
When it fails, you'd see something like 3 numbers with inequalities, and it is hard to figure out what failed, so I prefer keeping each assert with one comparison, so when it fails - it is clear what went wrong.

nvgoldin:
@gbenhaim - for some reason GitHub showed my replies twice, so I deleted one copy of them, and today it looks like all of them are gone :)
I hope you got them to your email, anyways updated V2 - thanks again for the review.

nvgoldin:
ci merge please

PR #641, opened 2017-07-18
Automation: Remove duplicate packages from mock *.packages

gbenhaim:
1. libguestfs-devel was twice on the list.
2. virtualenv, and tox are being installed before running the
   tests (using pip).
3. For some reason when installing tox from rpm, entry points aren't
   created, and then the job failed on 'tox: Command not found'. This
   patch will ensure that tox is always installed from PyPi.

Fixes: http://jenkins.ovirt.org/job/lago_master_github_check-patch-fc26-x86_64/216/console

### DIFF
--- a/automation/check-patch.packages
+++ b/automation/check-patch.packages
@@ -5,7 +5,6 @@ dnf-command(builddep)
 git
 grubby
 libffi-devel
-libguestfs-devel
 libvirt-devel
 libguestfs-devel
 libxslt-devel
@@ -13,6 +12,4 @@ openssl-devel
 pandoc
 python2-devel
 python2-dulwich
-python2-virtualenv
-python2-tox
 python-pip
--- a/automation/check-patch.packages.el7
+++ b/automation/check-patch.packages.el7
@@ -4,14 +4,11 @@ git
 libffi-devel
 libguestfs-devel
 libvirt-devel
-libguestfs-devel
 libxslt-devel
 openssl-devel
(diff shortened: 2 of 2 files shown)


gbenhaim:
ci test please

gbenhaim:
ci merge please

gbenhaim:
ci merge please

gbenhaim:
ci merge please

automation/check-patch.packages
```diff
@@ -5,14 +5,11 @@ dnf-command(builddep)
 git
 grubby
 libffi-devel
-libguestfs-devel
```
nvgoldin: Thanks, not sure how I managed to put it twice..

automation/check-patch.packages
```diff
@@ -5,14 +5,11 @@ dnf-command(builddep)
...
 libvirt-devel
 libguestfs-devel
 libxslt-devel
 openssl-devel
 pandoc
 python2-devel
 python2-dulwich
-python2-virtualenv
-python2-tox
```
nvgoldin: First, +1.
Second, `tox` is somewhat special because it requires a python interpreter(which obviously can't be installed via pip), I think tox's RPM should actually pull the available interpreters, but need to check that. As we currently only test 2.7, it doesn't matter, but in the future it will probably be better to install it from the RPM.
Could be that similar arguments apply to setuptools.

PR #646, opened 2017-07-26
cmd: drop group warning

nvgoldin:
Don't see any real usage for it anymore, especially when you can install
directly from pip without Lago group.


nvgoldin:
ci merge please

PR #650, opened 2017-08-20
Remove strange additions to network settings

mykaul:
We used to have way too many network definitions.
Per network, per network+interface, etc.

For now, only have domain and domain-ethX if X!= 0 (as eth0 is already 'domain')

Tested by running the complete suite of ovirt-system-tests.

### DIFF
--- a/lago/prefix.py
+++ b/lago/prefix.py
@@ -293,14 +293,9 @@ class Prefix(object):
         Populates the given net spec mapping entry with the nics of the given
         domain, by the following rules:
 
-            * If ``net`` is management, 'domain_name': nic_ip
-            * For each interface: 'domain_name-eth#': nic_ip, where # is the
-            index of the nic in the *domain* definition.
-            * For each interface: 'domain_name-net_name-#': nic_ip,
-            where # is a running number of interfaces from that network.
-            * For each interface: 'domain_name-net_name', which has an
-            identical IP to 'domain_name-net_name-0'
-
+            * If ``idx`` is 0, which is eth0, 'domain_name': nic_ip
+            * For each interface, but eth0: 'domain_name-eth#': nic_ip,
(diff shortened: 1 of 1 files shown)


mykaul:
@nvgoldin , @gbenhaim  - can you chek the failures? They are completely unrelated to this patch.

mykaul:
ci please test

mykaul:
ci test please

nvgoldin:
1. The failure is due to yapf:
```
17:16:10 lago/prefix.py
17:16:12 yapf diff: 
17:16:12 --- lago/prefix.py	(original)
17:16:12 +++ lago/prefix.py	(reformatted)
17:16:12 @@ -308,7 +308,7 @@
17:16:12          """
17:16:12          dom_name = dom['name']
17:16:12          idx = dom['nics'].index(nic)
17:16:12 -        if idx == 0: # eth0
17:16:12 +        if idx == 0:  # eth0
```

2. With regard to the naming the idea was:
  a. Always have domain-ethX - to ensure predictability. So you could always use the ethX DNS name.
  b. Create another alias for for 'domain', which is the same as the management network. 
This two I think are important - so you could always use a specific device by the DNS name, and ensure that the management one is the primary one(without -ethX)

For the other ones, the idea was more or less the same, but that you could map them by network name, i.e.:
domain-network_name-0
domain-network_name-1
I think it could be useful, though not sure if practical. @gbenhaim  - I remember we discussed this, but don't recall what was the motivation - you have an idea?

And last - why remove it? is there any performance improvements? I would guess adding DNS entries in dnsmasq should not cost much.

mykaul:
- dnsmasq was complaining on some names.
- the idea was nice, no one was using it.

mykaul:
Closing for the time being.

PR #656, opened 2017-09-13
collect-artifacts: Try to use Libguestfs

gbenhaim:
If an error occurred during SCP try to extract
logs using Libguestfs (if it's available).

### DIFF
--- a/lago/providers/libvirt/vm.py
+++ b/lago/providers/libvirt/vm.py
@@ -29,6 +29,7 @@ import sys
 import libvirt
 from lxml import etree as ET
 from textwrap import dedent
+from scp import SCPException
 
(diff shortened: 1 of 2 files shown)


lago/providers/libvirt/vm.py
```diff
@@ -341,7 +342,7 @@ def extract_paths(self, paths, ignore_nopath):
                 paths=paths,
                 ignore_nopath=ignore_nopath,
             )
-        except ExtractPathError as err:
+        except (ExtractPathError, SCPException) as err:
```
nvgoldin: the scp.py module can throw other exceptions(including IOError),
see: https://github.com/jbardin/scp.py/issues/82

PR #662, opened 2017-10-23
Set MTU for all interfaces to 9000.

mykaul:
This allows us to test with Jumbo frames.
Hopefully, it'll also generally improve performance
(NFS, iSCSI traffic, etc.)

I *think* it won't break if lower libvirt is used.
It won't set the MTU, and ifcfg-* value will be ignored.

### DIFF
--- a/lago/providers/libvirt/network.py
+++ b/lago/providers/libvirt/network.py
@@ -54,6 +54,12 @@ class Network(object):
     def gw(self):
         return self._spec.get('gw')
 
+    def mtu(self):
+        if self.libvirt_con.getLibVersion() > 3001001:
+            return self._spec.get('mtu', '1500')
+        else:
+            return '1500'
+
     def is_management(self):
         return self._spec.get('management', False)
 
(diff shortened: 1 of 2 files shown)


ifireball:
I think this should be an explicit option in the init file rather then the global default.

mykaul:
@ifireball - perhaps... Per network?
(BTW, right now it seems to be breaking host deploy!)

ifireball:
@mykaul yeah, probably per network, as storage network is likely to have jumbo frames, whil management may not

mykaul:
@ifireball - done, per network as requested.

ifireball:
I don't know enough about Lago internals to tell if this commit does what you say it does ;) I'll leave review up to @gbenhaim

mykaul:
@gbenhaim , @eedri - do you know if yapf style changed or something? It asks now for changes that are not part of my patch and are not consistent with what we've had in the past.

gbenhaim:
@mykaul I wonder if the `mtu` attribute should be added also to the bridge's xml https://libvirt.org/formatnetwork.html#elementsConnect

mykaul:
@gbenhaim - correct. In fact, apparently for 'nat' bridges, it sets it automatically for all NICs above, which is what we probably need.
Changed the code accordingly.

gbenhaim:
ci merge please

PR #675, opened 2017-11-26
sdk_tests: Check that VM is running

gbenhaim:
Add a test for checking if a VM is running.
Make sense to verify that the VMs are running before
testing ssh.

### DIFF
--- a/tests/functional-sdk/test_sdk_sanity.py
+++ b/tests/functional-sdk/test_sdk_sanity.py
@@ -171,6 +171,12 @@ def test_custom_gateway(vms, nets, init_dict):
             assert nets[net_name].gw() == net['gw']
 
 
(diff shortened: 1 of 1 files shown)


tests/functional-sdk/test_sdk_sanity.py
```diff
@@ -171,6 +171,12 @@ def test_custom_gateway(vms, nets, init_dict):
             assert nets[net_name].gw() == net['gw']
 
 
+def test_vm_is_running(vms, vm_name):
+    vm = vms[vm_name]
```
rgolangh: might be that the vm object is None. Check that

gbenhaim: How could it be? in which scenario?

gbenhaim:
ci merge please

PR #676, opened 2017-11-28
Switch the log of SSH retries to display

mykaul:
only if the 1st connection failed and we retry.
It's just annoying to see it on every SSH connection.


mykaul:
@gbenhaim - the FC26 failure has nothing to do with this patch (and we should probably drop some Fedora's from the tests). The failed test seems to be using vm-el7-3-base - time to move to el7-4 regardless, no?

mykaul:
ci test please

mykaul:
@gbenhaim - can we get this into 0.42?

gbenhaim:
ci merge please

PR #677, opened 2017-11-29
Wait 10 seconds before trying to SSH to VMs.

mykaul:
They are not up that quickly anyway.

### DIFF
--- a/lago/prefix.py
+++ b/lago/prefix.py
@@ -26,6 +26,7 @@ import os
 import shutil
 import subprocess
 from textwrap import dedent
+import time
 import urlparse
 import urllib
 import uuid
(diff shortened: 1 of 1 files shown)


mykaul:
@gbenhaim - can we get this into 0.42?

gbenhaim:
@mykaul It's not always true that `lago deploy` is called immediately after `lago start`, so I don't think that this option should be hardcoded.

What is wrong with `wait_for_ssh`?

mykaul:
When you have many VMs starting up (think performance suite of O-S-T), it tries to connect in parallel, to each one, giving up after 1s, they take 20-40s at least to respond.
1. It floods the logs.
2. The lago process is taking quite a bit of CPU.

At the end of the day, in 99% of the cases, the sleep is the right thing to do.
Worst case scenario, you've wasted 10s.

gbenhaim:
@mykaul https://github.com/lago-project/lago/pull/678/files please review.

PR #651, opened 2017-08-23
[WIP] clod-init: Adding cloud init support

gbenhaim:
1. ```user-data``` and ```meta-data``` can be specified directly under the ```cloud-init``` key in each 
       domain definition. The content can be specified explicitly or by adding path to a file.

2. For each domain, ISO file is generated and attached to the domain.

3. Default settings can be found in the default templates which later will be modified using Jinja.

4. Merge between user defined settings and default setting is done by utils.deep_update

Example:

```
domains:
  atomic-el:
    cloud-init:
      user-data:
        write_files:
          - path: /root/test
            content: bla_bla_bla
        users:
          - name: galito
        chpasswd:
          list: |
            galito:galgal
```

Signed-off-by: gbenhaim <galbh2@gmail.com>

PR #682, opened 2017-12-09
sysprep: Remove could-init

dominikholler:
This change improves the usage of cloud images in lago.
If we think the usage of [generic cloud images in OST](https://gerrit.ovirt.org/#/c/85170/) would be a good idea, this change would prepare this.

### DIFF
--- a/lago/templates/sysprep-base.j2
+++ b/lago/templates/sysprep-base.j2
@@ -10,3 +10,5 @@ upload {{ public_key }}:/root/.ssh/authorized_keys
 chmod 0600:/root/.ssh/authorized_keys
 run-command chown root:root /root/.ssh/authorized_keys
 {% endif -%}
+mkdir /etc/cloud
+touch /etc/cloud/cloud-init.disabled


dominikholler:
> I'd mask it instead, as it's faster.

I am not sure what means 'mask', but I disabled cloud-init in a simple and probably distribution  independent way.

mykaul:
delete /etc/systemd/system/multi-user.target.wants/kdump.service
delete /etc/systemd/system/multi-user.target.wants/postfix.service

Are two examples of masking. It is using systemd, but I assume all our
images do (?!)
Y.

dominikholler:
> delete /etc/systemd/system/multi-user.target.wants/kdump.service
> delete /etc/systemd/system/multi-user.target.wants/postfix.service
> 
> Are two examples of masking.


This way of disabling are used for specif el or fc versions.
The change proposed here would affect all linux distributions.

> It is using systemd, but I assume all our
> images do (?!)

I see no reason here to have a dependency to systemd, even it would not
hurt.

Additional there are 4 *.service links for cloud-init
in /etc/systemd/system/multi-user.target.wants and I have no idea if
they depend on cloud-init version.

Because the [cloud-init
documentation](http://cloudinit.readthedocs.io/en/latest/topics/boot.html#generator)
suggest to disable cloud-init by the /etc/cloud/cloud-init.disabled or
kernel command line, I think we should disable cloud-init like this.

ifireball:
I'd really rather someone goes ahead and implements full cloud-init support for Lago rather then workarounds. I think various pieces of this work had already been done in the past.

gbenhaim:
1. There is a WIP of adding cloud-init support to lago: https://github.com/lago-project/lago/pull/651 
    (actually its almost ready, just need to add some unit and functional tests).

2. I don't think that disabling cloud-init should be hard coded.

3. We can use the `build` section in lago's init file in order to disable cloud-init. http://lago.readthedocs.io/en/latest/BUILD.html

4. The following list of symlinks can be used in order to disable cloud-init:

```
link /dev/null:/etc/systemd/system/cloud-init.service
link /dev/null:/etc/systemd/system/cloud-init-local.service
link /dev/null:/etc/systemd/system/cloud-config.service
link /dev/null:/etc/systemd/system/cloud-final.service
```

dominikholler:
> 3. We can use the build section in lago's init file in order to disable cloud-init.
> http://lago.readthedocs.io/en/latest/BUILD.html

This is great! This way we can disable could-init per image and this makes this change obsolete.

Thanks for the discussion, I will go on in https://gerrit.ovirt.org/#/c/85170/, because no change to lago seems required.

PR #686, opened 2017-12-22
Remove mtu=1500 from ifcfg files.

mykaul:
Not sure why I've added it when I added jumbo frames support to
Lago. It actually makes not sense - if the network is set for
9000b for example (and virtio with newer kernel detects it
automatically), the MTU should not be set via ifcfg.

Removed it and tested with ovirt-system-tests with both
default MTU and 9000b and both work well.


gbenhaim:
ci merge please

Issue #685, opened 2017-12-19
sdk: Expose the path to lago.log


gbenhaim:
This way it will be easier to extract the log from the env.

gbenhaim:
https://github.com/lago-project/lago/commit/6eda019d8afc0ada1781b349fde7cbf660de76bc

PR #687, opened 2017-12-28
prefix: Expose Paths object in the SDK

gbenhaim:
- Expose Paths object in the SDK (currently, we can only expose methods,
  not properties).

- Expose the paths attribute as a property (for internal use inside lago).

- Added some documentation about the Paths class.

- In Prefix class, change the attribute name that contains the prefix path from
  "prefix" to "prefix_path" (the convention in our code is that a variable
  named "prefix" is a Prefix object).

- Remove "_prefix" attribute from the prefix object. This attribute
  used to hold the path to the prefix, but the same value is also stored
  in the Paths object, so there is no reason to save it twice.

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -227,12 +227,13 @@ def do_cleanup(prefix, **kwargs):
 def do_destroy(
     prefix, yes, all_prefixes, parent_workdir, prefix_name, **kwargs
 ):
-    warn_message = prefix.paths.prefix
-    path = prefix.paths.prefix
 
     if all_prefixes:
         warn_message = 'all the prefixes under ' + parent_workdir.path
         path = parent_workdir.path
+    else:
+        warn_message = prefix.paths.prefix_path()
+        path = warn_message
 
     if not yes:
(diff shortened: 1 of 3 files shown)


gbenhaim:
@nvgoldin 
Any simple idea on how to expose `properties` of the prefix in the SDK?
Using both `expose` and `property` decorators don't work.

nvgoldin:
Sorry, missed this. Will look. In principal I didn't test it with properties, I assume it has to do with the fact the properties implement them self the descriptor protocol which we override one if its methods. Btw, you tried changing the order of the decorators?

gbenhaim:
closes https://github.com/lago-project/lago/issues/685

gbenhaim:
@nvgoldin yes I've tried. If the property decorator goes first,  sdk_utils.expose raise an exception because of functools.wraps.

gbenhaim:
ci test please

gbenhaim:
ci test please

gbenhaim:
ci merge please

PR #688, opened 2017-12-28
style: Control manually on yapf version

gbenhaim:
- It will save us to format the entire code base every couple
  of month when a new version of yapf is released.

- Pin yapf to version 0.20


gbenhaim:
ci merge please

PR #696, opened 2018-02-25
Move machine type to Q35.

mykaul:
I think it's time, and it'll enable us to add PCIE support.
Later, hopefully we'll be able to do device-passthrough.

I've also took the opportunity to get rid of the video device.
It's not needed.


mykaul:
This requires NUMA, which is OK - unless we are using a single vCPU. We could add NUMA awareness to a single CPU as well, will probably make life easier.

PR #698, opened 2018-03-15
Move to download images using requests library.

mykaul:
More modern, hopefully somewhat faster.


eedri:
Is python-requests available on RHEL and Centos? this will be needed for lago demo tool and running lago on RHEL.

mykaul:
@eedri I'd be very surprised if it's not, but apparently things are failing :(

requirements.txt
```diff
@@ -15,3 +15,4 @@ wrapt
 Jinja2
 xmltodict
 netaddr
+requests
```
gbenhaim: It should be added to the spec file as well.

lago/templates.py
```diff
@@ -24,9 +24,10 @@
 import logging
 import os
 import posixpath
+import requests
 import shutil
-import urllib
 import sys
+import urllib
```
gbenhaim: Why do we still need `urllib`?

mykaul: It is still used in the code.

PR #705, opened 2018-04-09
Reuse libvirt connections

mykaul:
1. For some reason, we've opened many connections to libvirt.
A single libvirt connection should suffice.
2. Every connection (for VM) would do a getCapabilities and
getVersion - for no reason - reuse from the same one.
3. We've never really bothered to try and close the connections.
I don't think I've closed all connections, but I've made an honest
effort.
4. As consequence, we do not generate the CPU (and NUMA config) so many times (this is what started this whole patch).

### DIFF
--- a/lago/providers/libvirt/vm.py
+++ b/lago/providers/libvirt/vm.py
@@ -56,24 +56,23 @@ class LocalLibvirtVMProvider(vm_plugin.VMProviderPlugin):
         self._has_guestfs = 'lago.guestfs_tools' in sys.modules
         libvirt_url = config.get('libvirt_url')
         self.libvirt_con = libvirt_utils.get_libvirt_connection(
-            name=self.vm.virt_env.uuid + libvirt_url,
             libvirt_url=libvirt_url,
         )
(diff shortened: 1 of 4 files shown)


mykaul:
@gbenhaim - ping? This also nicely reduces lago logs to become more readable.

lago/providers/libvirt/vm.py
```diff
@@ -455,7 +458,8 @@ def cpu_model(self):
             str: CPU model
 
         """
-        return self._cpu.model
+        self.cpu()
+        return self.cpu.model
```
gbenhaim: `cpu` is a function, not a property.

lago/virt.py
```diff
@@ -86,7 +86,6 @@ def __init__(self, prefix, vm_specs, net_specs):
 
         libvirt_url = config.get('libvirt_url')
         self.libvirt_con = libvirt_utils.get_libvirt_connection(
```
gbenhaim: I think that the `virt` object shouldn't be aware of provider-specific details, thus the `libvirt_con` reference (and all the code that use it) can be removed from `virt.py`.

PR #710, opened 2018-04-24
Remove one call to get_ssh_client()

mykaul:
Since in ssh_reachable() we already get a SSH client connection,
let's save it in the (unused so far) _ssh_client var.
Then reuse it, in _scp() command.


mykaul:
ci test please

mykaul:
ci test please

mykaul:
ci please test

mykaul:
ci test please

lago/plugins/vm.py
```diff
@@ -686,19 +685,23 @@ def _normalize_spec(cls, spec):
...
-        client = ssh.get_ssh_client(
-            propagate_fail=propagate_fail,
-            ip_addr=self.ip(),
-            host_name=self.name(),
-            ssh_key=self.virt_env.prefix.paths.ssh_id_rsa(),
-            username=self._spec.get('ssh-user'),
-            password=self._spec.get('ssh-password'),
-        )
+        if self._ssh_client is not None:
+            client = self._ssh_client
```
gbenhaim: Can we use the same client in multiple threads?

gbenhaim: I think that it would be better to add `try/catch` when trying to get the client in `_scp`.
It will prevent a case where:
1. You check if the vm is ssh reachable
2. Something makes the vm go offline
3. You call the `_scp` context manager.

mykaul: What am I supposed to do in this case? What do I do in the catch? Nothing I can do then if the VM is offline - we'll fail miserably anyway.

gbenhaim: 1. You can fall back to use `libguestfs`.
2. Print an explanation about the failure. It's better than to see the entire stack trace on screen.

PR #712, opened 2018-05-01
Reduce the number of DHCP entries we add to libvirt.

mykaul:
We only need really one entry per host.
The fact a host has many DNS names (X-eth..., X-<network-name>, etc...)
doesn't mean we need so many DHCP entries.

### DIFF
--- a/lago/providers/libvirt/network.py
+++ b/lago/providers/libvirt/network.py
@@ -226,7 +226,13 @@ class NATNetwork(Network):
                 )
             )
 
-            for hostname, ip4 in self._spec['mapping'].items():
+            ipv4s = []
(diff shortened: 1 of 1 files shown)


mykaul:
@ifireball - the CI failure does not seem to be related to this patch:
libvirtError: Failed to find group record for gid '1001'
12:29:00 # ---

Anything wrong in the CI infra?

mykaul:
ci test please

gbenhaim:
We invoke the CLI functional tests with lago's group, it means that the group should exist on the Jenkins slave (which in this case wasn't true). I wonder if we really need to invoke the tests with lago's group.

gbenhaim:
ci test please

gbenhaim:
ci merge please

PR #719, opened 2018-05-09
Don't use check_alive decorator in ssh_reachable

gbenhaim:
check_alive decorator raise RuntimeError if the vm is not running.
The API of ssh_reachable defines that "False" should be returned if the
vm is not ssh reachable (which is the case when the vm is not running).


mykaul:
LGTM, especially that there's a single caller (I think) to ssh_reachable() and it is stesting first that's the host is alive.

PR #720, opened 2018-05-09
Update get ssh client

gbenhaim:
Use custom exception in "get_ssh_client"

    If there was a failure to get a connected ssh client, raise
    "LagoSSHTimeoutException". This change will make it safer to catch
    exceptions raised by "get_ssh_client".

Add docstrings to "get_ssh_client"


gbenhaim:
ci test please

gbenhaim:
I want to use try/catch with `LagoSSHTimeoutException` in  `lago.plugins.vm.extract_paths` instead of calling `vm.ssh_reachable`. The current implementation can introduce a race condition.

gbenhaim:
ci merge please

Issue #631, opened 2017-07-05
log collector could not get logs (timed out)


gilShin:
Looking at the logs what happened was:
1. Lago checked the engine was SSH reachable - this was true.
2. Then it tried connecting via SSH and collect the logs and timed out.

On (1) we have retries and guards, on (2) we don't, as we assume (1)
just passed. I guess in some conditions that logic can be flawed

gbenhaim:
Can you add thelink to the build that failed?

gilShin:
Sורק:
http://jenkins.ovirt.org/job/test-repo_ovirt_experimental_master/7459/

PR #711, opened 2018-04-24
libvirt calls cleanup

mykaul:
Less calls (hopefully) to libvirt, more specific lookups for VMs
and networks.
A bit of cleanup - remove of libvirt related stuff from lago/virt.py

PR #724, opened 2018-05-29
Use try except in "extract_paths"

gbenhaim:
- Use try/except instead of explicitly checking if the VM is running and
  available via SSH.

- Add "check_running" decorator to "_extract_paths_scp".
  There is no reason to try and use SCP if the VM isn't running.

- Change "check_running" decorator to raise a custom exception, instead
  of a "RuntimeError". The custom exception will also be formatted
  nicely on stdout (the stack trace will be dumped to the log, while
  the message will be printed to stdout).


gbenhaim:
closes https://github.com/lago-project/lago/issues/631

gbenhaim:
@mykaul thoughts?

mykaul:
I didn't look at the patch thoroughly, but I have nothing against a try/catch. I did not encounter the false-positive 'VM is not running' issue since I've changed the libvirt calls (https://github.com/lago-project/lago/pull/711 ) though.

gbenhaim:
ci merge please

Issue #492, opened 2017-03-21
vm.py's defined() function is looking if a domain is defined, not if it's running


mykaul:
Code:
```
    def defined(self):
        dom_names = [dom.name() for dom in self.libvirt_con.listAllDomains()]
        return self._libvirt_name() in dom_names
````

1. It's inefficient - you want to look for a specific domain, do it using lookup
2. It's not checking if it's running, but if it's defined (currently, we don't have persistent domains, but we could have).

See http://libvirt.org/docs/libvirt-appdev-guide-python/en-US/html/libvirt_application_development_guide_using_python-Guest_Domains-Information-IsActive.html for example.

gbenhaim:
2 was fixed in https://github.com/lago-project/lago/pull/711

PR #726, opened 2018-05-31
Ask the provider if the vm is running

gbenhaim:
#### Ask the provider if the VM is running

We ask the provider for the state of the VM, so it makes
sense to also ask the provider if the VM is running.

#### Improve VM state query
- Don't list all the VMs, just lookup the VM that is associated
  with the provider.

- Handle edge cases where an unknown state was returned from Libvirt,
  or when the VM doesn't exist.

closes https://github.com/lago-project/lago/issues/492


gbenhaim:
ci merge please

PR #742, opened 2018-06-19
Update developers docs

galitf:
updated documentation flow


galitf:
has a new squash patch

PR #751, opened 2018-06-26
show message when prefix already exists

galitf:
Added a friendly message when prefix already exists

### DIFF
--- a/lago/workdir.py
+++ b/lago/workdir.py
@@ -268,15 +268,10 @@ class Workdir(object):
             The newly created prefix
 
         Raises:
-            PrefixAlreadyExists: if the prefix name already exists in the
+            LagoPrefixAlreadyExistsError: if prefix name already exists in the
                 workdir
         """
         if os.path.exists(self.join(name)):
-#            raise PrefixAlreadyExists(
-#               'Prefix with name %s already exists in workdir %s' %
-#                (name, self.path)
-#            )
-
             raise LagoPrefixAlreadyExistsError(name, self.path)
 
         self.prefixes[name] = self.prefix_class(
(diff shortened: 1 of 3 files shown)


gbenhaim:
ci test please

lago/workdir.py
```diff
@@ -266,14 +268,11 @@ def add_prefix(self, name, *args, **kwargs):
             The newly created prefix
 
         Raises:
-            PrefixAlreadyExists: if the prefix name already exists in the
+            LagoPrefixAlreadyExistsError: if prefix name already exists in the
                 workdir
         """
         if os.path.exists(self.join(name)):
-            raise PrefixAlreadyExists(
```
gbenhaim: Don't you need to delete the class of this exception?

lago/workdir.py
```diff
@@ -503,3 +502,18 @@ def set_current(prefix_name, parent_workdir, **kwargs):
...
     parent_workdir.set_current(new_current=prefix_name)
+
+
+class LagoPrefixAlreadyExistsError(utils.LagoException):
+    def __init__(self, name, path):
+        super().__init__(
+            dedent(
+                """
+                Failed Prefix name {} already exists in workdir {}
+                1) Destroy old lago VM in order to create new VM
```
gbenhaim: I don't think that we should specify command names and flags here, because if the name of the command (or its flags) will change, we will need to remember change this message.

A good example is the `WORKDIR-PATH` positional parameter that we think to remove.

galitf:
@gbenhaim ready for review

galitf:
Created a new patch, Fix in #779

PR #752, opened 2018-06-26
 added friendly message when a file to copy does not exist in copy-from-vm copy-to-vm

galitf:
Added a friendly message when a file to be copy doesn't exists 
updated in both functions:copy-from-vm copy-to-vm

### DIFF
--- a/lago/plugins/vm.py
+++ b/lago/plugins/vm.py
@@ -289,7 +289,7 @@ class VMProviderPlugin(plugins.Plugin):
                     '%s: SCPException _extract_paths_scp message', err.args[0]
                 )
                 if not ignore_nopath:
-                    raise LagoCopyFilesFromVMError(remote_path, local_path)
+                    raise LagoCopyFilesFromVMError(host_path, guest_path)
 
 
 class VMPlugin(plugins.Plugin):
--- a/lago/plugins/vm.py
+++ b/lago/plugins/vm.py
@@ -285,16 +285,8 @@ class VMProviderPlugin(plugins.Plugin):
                     propagate_fail=False
                 )
             except SCPException as err:
-                err_substr = ': No such file or directory'
-                if len(err.args) > 0 and isinstance(
-                    err.args[0], basestring
-                ) and err_substr in err.args[0]:
-                    if ignore_nopath:
-                        LOGGER.debug('%s: ignoring', err.args[0])
-                    else:
-                        raise ExtractPathNoPathError(err.args[0])
-                else:
-                    raise
+                if not ignore_nopath:
+                    raise LagoCopyFilesFromVMError(remote_path, local_path)
 
 
 class VMPlugin(plugins.Plugin):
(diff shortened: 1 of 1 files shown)


lago/plugins/vm.py
```diff
@@ -67,6 +66,23 @@ def __init__(self, vm_name):
...
+            'Failed, Copy files/directory {} does not exist!'.
+            format(local_files)
+        )
+
+
+class LagoCopyFilesFromVMError(utils.LagoUserException):
+    def __init__(self, remote_files, local_files):
+        super().__init__(
+            'Failed, Copy files/directory from {} does not exist!'.format(
+                remote_files, local_files
```
gbenhaim: The string has only 1 place for a variable, and you are passing 2 variables.

lago/plugins/vm.py
```diff
@@ -411,22 +423,39 @@ def copy_to(self, local_path, remote_path, recursive=True):
...
         self, remote_path, local_path, recursive=True, propagate_fail=True
     ):
-        with self._scp(propagate_fail=propagate_fail) as scp:
-            scp.get(
-                recursive=recursive,
-                remote_path=remote_path,
-                local_path=local_path,
-            )
+        with LogTask(
+            'Copy from %s:%s to %s' % (self.name(), remote_path, local_path),
```
gbenhaim: The commit message doesn't specify that you added this log message.

lago/plugins/vm.py
```diff
@@ -411,22 +423,39 @@ def copy_to(self, local_path, remote_path, recursive=True):
...
+                        recursive=recursive,
+                        remote_path=remote_path,
+                        local_path=local_path,
+                    )
+            except SCPException as err:
+                err_substr = ': No such file or directory'
+                if len(err.args) > 0 and isinstance(
+                    err.args[0], basestring
+                ) and err_substr in err.args[0]:
+                    if propagate_fail:
```
gbenhaim: `propagate_fail` is a variable should be used only by the logger.
You can see how it's being used in `lago.log_utils:604`.

In our case, I think that we can just raise the exception.

lago/plugins/vm.py
```diff
@@ -411,22 +423,39 @@ def copy_to(self, local_path, remote_path, recursive=True):
...
+                        LOGGER.debug('%s: ignoring', err.args[0])
+                        raise LagoCopyFilesFromVMError(remote_path, local_path)
```
gbenhaim: Maybe let's add the original message to the output that is printed to the user, what do you think?

lago/plugins/vm.py
```diff
@@ -411,22 +423,39 @@ def copy_to(self, local_path, remote_path, recursive=True):
...
-                    recursive=recursive,
-                )
+            try:
+                with self._scp() as scp:
+                    scp.put(
+                        files=local_path,
+                        remote_path=remote_path,
+                        recursive=recursive,
+                    )
+            except OSError:
```
gbenhaim: This will catch other exceptions as well (such as missing permissions), which is OK.
Yet, we need to change `LagoCopyFilesToVMError`, because its message refers only to file which don't exist. I think that we can append the original error to the message that we output for the user.

```python
except OSError as e:
    raise(local_path, str(e))
```

lago/plugins/vm.py
```diff
@@ -268,17 +284,13 @@ def _extract_paths_scp(self, paths, ignore_nopath):
...
             except SCPException as err:
-                err_substr = ': No such file or directory'
-                if len(err.args) > 0 and isinstance(
-                    err.args[0], basestring
-                ) and err_substr in err.args[0]:
-                    if ignore_nopath:
-                        LOGGER.debug('%s: ignoring', err.args[0])
-                    else:
-                        raise ExtractPathNoPathError(err.args[0])
+                if ignore_nopath:
```
gbenhaim: This `if` block should be under the first except, because it's the one which handles files the doesn't exist.

galitf: If I move the block to the previous except it will fails.
It provides an exception on:
SCPException: scp: /var/log/something.log: No such file or directory
The ignore path in this case is related to the scp how to perform

galitf:
@gbenhaim please review

PR #761, opened 2018-08-30
ansible: Add special case for the group key

gbenhaim:
Some ansible playbooks expect that a host will be part
of a certain group. The current implementation doesn't allow to
specify exactly what will be the name of a group in the inventory,
since it adds group in the format of "key=value".

With this patch, each VM that has the "groups" list in its spec, will
be also a part of each group in that list.


gbenhaim:
ci merge please

gbenhaim:
ci merge please

gbenhaim:
ci merge please

PR #767, opened 2018-10-07
Improve error reporting

gbenhaim:
#### Print thread's exception info to the log

This change will improve the way we show errors to the user.
Most of the times, there is no need to expose the low level details of
an exception to the user. If needed, the information can be printed
explicitly.

#### Adding LagoDeployError

Use a custom exception if the invocation of the deploy scripts fails.
This change will hide the traceback from the user, while keeping
the stderr of the script.

#### Print exception message with logger.error

Error messages should be printed with the ERROR log level.

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -968,7 +968,7 @@ def main():
     try:
         cli_plugins[args.verb].do_run(args)
     except utils.LagoException as e:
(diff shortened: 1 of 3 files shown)


gbenhaim:
ci merge please

PR #769, opened 2018-11-18
Fix build errors

gbenhaim:
#### Use a fixed version of Falke8

This change will eliminate CI failures due to a change
in Flake8 version.

Flake8 version update should be manageable.

#### Set locale when building docs

Without setting the locale the build is failing with:

```
locale.setlocale(locale.LC_ALL, '')
locale.Error: unsupported locale setting
```

####  Ignore deprecation error caused by stevedore

setuptools (which is used by stevedore) 11.3 deprecated the required flag
to load() in favor of two new methods.

A fix was introduced in stevedore 1.2.0.

Since we are using and older version of stevedore (0.14) from epel,
lago outputs the deprecation warning, which isn't helpful for our users.

A long term solution will be to use a newer version of stevedore, but
currently it only available in Openstack repos.

#### Use PEM key format for ssh

OpenSSH-7.8 introudced a new key format which isn't supported by
paramiko [1]. Use PEM key format instead.

We have not seen this error in CI since the fallback is to use password
authentication.

####  tests: Use retries when running systemd-analyze

"test_systemd_analyze" sporadically fails on
"Bootup is not yet finished. Please try again later".

Use retries in order to mitigate this issue.

### DIFF
--- a/lago/plugins/__init__.py
+++ b/lago/plugins/__init__.py
@@ -22,6 +22,7 @@
 """
 from stevedore import ExtensionManager
 import logging
+import warnings
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


gbenhaim:
ci test please

gbenhaim:
ci test please

gbenhaim:
ci merge please

gbenhaim:
ci merge please

PR #771, opened 2018-11-22
Change file locking method to flock

gbenhaim:
The current locking method leaves stale lock when Lago
is being terminated using SIGKILL.
Using flock we can overcome this issue.

Also:

1. Make FileLock context manager more generic.
2. Add a test.
3. Adjust templates.py - we don't use the "_lock" decorator.


gbenhaim:
ci test please

gbenhaim:
ci merge please

PR #775, opened 2018-11-29
Meaningful message when there is Unknown VM type

galitf:
Signed-off-by: Galit <grosenth@redhat.com>

### DIFF
--- a/lago/virt.py
+++ b/lago/virt.py
@@ -26,6 +26,7 @@ import os
 import uuid
 
 import yaml
+from future.builtins import super
 
 from lago import log_utils, plugins, utils
 from lago.config import config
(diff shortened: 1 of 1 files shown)


lago/virt.py
```diff
@@ -36,6 +37,14 @@
 log_task = functools.partial(log_utils.log_task, logger=LOGGER)
 
 
+class LagoUnknownVMTypeError(utils.LagoUserException):
+    def __init__(self, vm_type_name, vm_types):
+        super().__init__(
+            'Unknown VM type: {0}, available types: {1}, \
+            need to install lago-ovirt plugin'.format(vm_type_name, vm_types)
```
gbenhaim: Didn't we agree to not mention `lago-ovirt` plugin, since it's possible that the `vm_type` will come from a different plugin?

gbenhaim:
ci merge please

PR #779, opened 2018-12-03
Show message when prefix already exists

galitf:
Signed-off-by: Galit <grosenth@redhat.com>

### DIFF
--- a/lago/workdir.py
+++ b/lago/workdir.py
@@ -30,6 +30,8 @@ import os
 import logging
 import shutil
 from functools import partial, wraps
+from future.builtins import super
+from textwrap import dedent
 
 from . import (prefix, utils)
 from .plugins import cli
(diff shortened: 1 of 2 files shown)


lago/workdir.py
```diff
@@ -503,3 +502,16 @@ def set_current(prefix_name, parent_workdir, **kwargs):
...
     """
     parent_workdir.set_current(new_current=prefix_name)
+
+
+class LagoPrefixAlreadyExistsError(utils.LagoException):
+    def __init__(self, name, path):
+        super().__init__(
+            dedent(
+                """
+                Failed Prefix name {} already exists in workdir {}
```
gbenhaim: I think that we should keep the old error message.
This error message is wrong because creating a prefix isn't relate to running a VM.
Running a VM occurs in `lago start`, creating a prefix occurs in `lago init`.

PR #780, opened 2018-12-09
version_manager: Support shallow git clones

gbenhaim:
When creating a shallow clone of a repo, the history is being
truncated. This behaviour cause the first commit in the repo to have
a parent, but the metadata of parent isn't available.

We need this change since "read the docs" moved to use shallow git
clones.


gbenhaim:
ci merge please

PR #791, opened 2019-03-10
Set file permissions on lock files

gbenhaim:
Explicitly set the permissions of lock files, thus even if umask other
then '0002' is configured, different users in the lago group can try to
acquire the lock.


galitf:
ci merge please

PR #793, opened 2019-03-14
do not merge: Use fixed scp version in the func test

gbenhaim:
I suspect that scp==0.13.1 introduce a bug that cause "test_collect_exists"
to fail with the following error:

self = <scp.SCPClient object at 0x7fa27813c690>, cmd = ('',)

    def _recv_popd(self, *cmd):
>       assert self._depth > 0
E       AssertionError


gbenhaim:
scp 0.13.2 was released with a fix.

PR #795, opened 2019-03-31
out-format: Always print yaml in block style

gbenhaim:
- Block style is more readable.
- Setting the style explicitly in Lago will make the output
  deterministic (not dependent on python yaml defaults).


gbenhaim:
ci merge please

PR #812, opened 2020-05-29
Fixes in 'prefix' module

tinez:
This PR adds a fix for libvirt 6 that requires backing file formats to be specified explicitly and support for relative backing file paths to lago.


michalskrivanek:
LGTM

PR #813, opened 2020-06-29
Add '--ssh-key' option to 'lago init'

tinez:
This PR adds a '--ssh-key' option to 'lago init' command. This makes possible using ssh keys which are already injected into VM. On top of that, there are two small py38 fixes.

### DIFF
--- a/lago/cmd.py
+++ b/lago/cmd.py
@@ -113,6 +113,14 @@ in_lago_prefix = in_prefix(
         'section of all the disks'
     ),
 )
+@lago.plugins.cli.cli_plugin_add_argument(
+    '--ssh-key',
+    action='store',
+    help=(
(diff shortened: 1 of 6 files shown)


michalskrivanek:
LGTM

tinez:
Verified on my server

tinez:
ci merge please

nirs:
Looks like ost needs this patch, and it works. Can we merge this?

tinez:
> Looks like ost needs this patch, and it works. Can we merge this?

I'm not able to. The "ci please merge" request always fails, because the pipelines are broken (and won't be fixed). That's why currently the version of lago used for OST comes from my private copr.

tinez:
ci merge please

PR #819, opened 2021-01-19
utils: Fix Flock for multi-user scenarios

tinez:
When trying to use `Flock` class on files that are not
owned by the user, but over which the user has control
by belonging to a specific group, the call to `open` fails.

This results in errors for i.e. subnet leasing like:

```
 Failed to acquire a lock for store /var/lib/lago/subnets.
 This failure can be caused by several reasons:
 1. Another 'lago' environment is using the store.
 2. A stale lock was left in the store.
 3. You don't have R/W permissions to the store.
```

This patch fixes the issue by using the lower-level `os.open` call
instead with `0o660` mode specified.

It's not trivial to implement a test for this in CI, but the code
can be tested by multiple users running an example script of:

```
 import time
 from lago import utils

 lock = utils.Flock("/var/lib/lago/subnets/subnet-lease.lock")

 print("trying to acquire...")
 lock.acquire()
 print("acquired, waiting for keyboard interrupt...")

 try:
     time.sleep(100000000)
 except KeyboardInterrupt:
     pass

 print("releasing...")
 lock.release()
 print("done")
```

Signed-off-by: Marcin Sobczyk <msobczyk@redhat.com>


tinez:
ci test

tinez:
ci test

